PR #964, opened 2026-06-30
[19635] fix: Add scope="col" to th elements for td-has-header a11y compliance

geromegrignon:
fixes: #3 

Description
Adds scope="col" attribute to <th> elements in Table and TreeTable components to fix the td-has-header accessibility violation.

Changes

- Added updateThScope() method that sets scope="col" on all <th> elements
- Method is called on view init, when data loads, and when columns change
- Ensures dynamically loaded content and toggled columns maintain accessibility compliance

Why
Without the scope attribute, screen reader users cannot understand data context when navigating table cells. The scope="col" attribute creates a programmatic relationship between data cells and their column headers.

Testing

1. Run axe DevTools or Lighthouse accessibility audit on Table/TreeTable demos
2. Verify td-has-header rule passes
3. Inspect <th> elements and confirm scope="col" is present

References

- axe rule: td-has-header
- WCAG: 1.3.1 Info and Relationships (Level A)

> Migrated from `primefaces/primeng#19636` — originally by `@fkwq34` on 2026-06-23

---
*Original base: `master` @ `8c0ce70`, head: `tables-accessibility-scopes` @ `35ad53b`*

### DIFF
--- a/packages/primeng/src/stepper/stepper.ts
+++ b/packages/primeng/src/stepper/stepper.ts
@@ -80,8 +80,7 @@ export interface StepPanelContentTemplateContext {
     changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
     host: {
-        '[class]': 'cx("root")',
-        '[attr.role]': '"presentation"'
+        '[class]': 'cx("root")'
     },
     providers: [StepListStyle, { provide: STEPLIST_INSTANCE, useExisting: StepList }, { provide: PARENT_INSTANCE, useExisting: StepList }],
     hostDirectives: [Bind]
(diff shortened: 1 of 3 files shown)


geromegrignon:
*[Originally posted by `@rosenthalj` on 2026-06-24T08:20:10.000Z]*

I only quickly reviewed the changes you made to table.ts.  Wouldn't it be more efficient to be proactive vs reactive by adding the missing attribute when the "**th**" is created?  How does querying/setTimeout effect performance when a table has a large number of columns or a large number of rows? 

What would happen if a table (HTML and/or PrimeNG) was implemented within a PrimeNG table?  PrimeNG  (and your change) should not be modifying any embedded non-PrimeNG tables.

geromegrignon:
*[Originally posted by `@fkwq34` on 2026-06-24T18:13:10.000Z]*

`@rosenthalj` Good points. I have reworked the approach:

- Switched to a MutationObserver on the <thead> instead of running querySelectorAll on every change — it now only fires when th elements are added
- Scoped the selector to :scope > tr > th so we don't accidentally change any nested tables embed

Let me know if you see any other issues.

geromegrignon:
Closed as too old.

PR #981, opened 2026-06-30
[19580] fix: Remove hardcoded aria-labels in date picker

geromegrignon:
fixes: #24

Fixes an accessibility label-content-name-mismatch issue in Datepicker header navigation.

The month and year switch buttons were using static aria-label values (chooseMonth / chooseYear) that could differ from their visible labels (e.g. January, 2026). This caused a mismatch between visible text and accessible name, impacting speech-input users and triggering axe violations.

This PR removes those explicit aria-label bindings so the buttons use their visible text as the accessible name, aligning behavior with WCAG/axe expectations while keeping existing Datepicker functionality unchanged.

> Migrated from `primefaces/primeng#19582` — originally by `@fkwq34` on 2026-05-14

---
*Original base: `master` @ `6da50ba`, head: `datepicker-wcag-issue` @ `fe4f181`*


geromegrignon:
Closed as too old.

PR #997, opened 2026-06-30
fix: #19531 - tooltip respects overlayAppendTo config

geromegrignon:
## Description

Fixes [#48](https://github.com/primefaces/primeng/issues/48).

The `pTooltip` directive exposed a `$appendTo` computed signal that resolved `config.overlayAppendTo()`, but the signal was never consumed. The three DOM paths — `create()`, `getHostOffset()`, `remove()` — read `_tooltipOptions.appendTo` directly (default `'body'`), so the global `PrimeNG.overlayAppendTo` configuration was ignored. This broke overlay portaling in Shadow DOM / Angular Elements scenarios, where Tooltip diverged from other overlays (Select, DatePicker, MultiSelect, etc.) that correctly use the global config.

## Changes

**`packages/primeng/src/tooltip/tooltip.ts`**
- Extended the `$appendTo` computed signal with a full resolution chain:
  1. local `appendTo` input
  2. `config.overlayAppendTo()` (skipped when it is the default `'self'`, so the historical `'body'` default is preserved for non-configured users)
  3. `_tooltipOptions.appendTo` (`'body'` by default, preserves `tooltipOptions.appendTo` back-compat)
- Routed `create()`, `getHostOffset()` and `remove()` through `this.$appendTo()` instead of `getOption('appendTo')`, ensuring create/remove use the same container.
- Removed a dead `onChanges` branch for `simpleChange.appendTo` — `appendTo` is an `input()` signal and does not emit `SimpleChanges`; in Angular 21 the block still fired with `currentValue === undefined` and wiped the `'body'` default.

## Tests

Added a new `describe('Append To')` block in `tooltip.spec.ts` with 5 regression tests covering:
- default append to `document.body`;
- `config.overlayAppendTo === 'self'` (default) does not hijack the tooltip;
- `config.overlayAppendTo` set to a custom element is honored;
- local `[appendTo]` input wins over the global config;
- `remove()` cleans up from the same container used by `create()`.

All tooltip unit tests pass (`49/49` via `ng test primeng --include='**/tooltip/tooltip.spec.ts'`).

## Backwards compatibility

No breaking change:
- users without any configuration keep the historical `'body'` behavior;
- `tooltipOptions: { appendTo: ... }` continues to work;
- `[appendTo]` input binding continues to work and keeps the highest priority.

> Migrated from `primefaces/primeng#19542` — originally by `@vitopio-prete` on 2026-04-23

---
*Original base: `master` @ `451ca15`, head: `fix/tooltip-overlay-append-to` @ `60b64bb`*

### DIFF
--- a/packages/primeng/src/tooltip/tooltip.spec.ts
+++ b/packages/primeng/src/tooltip/tooltip.spec.ts
@@ -3,6 +3,7 @@ import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 
 import { TooltipOptions } from 'primeng/api';
+import { PrimeNG } from 'primeng/config';
 import { Tooltip } from './tooltip';
 
 @Component({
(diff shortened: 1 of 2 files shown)


geromegrignon:
Closed as too old.

PR #1000, opened 2026-06-30
fix: #19485 - Table virtualScroll rows not visible with percentage scrollHeight

geromegrignon:
### Description  
This PR fixes a regression in PrimeNG v21 where p-table rows are not visible when virtualScroll is enabled with percentage-based scrollHeight (for example 100%).  
Root cause: the virtual scroller viewport could resolve against an auto-height chain, causing the effective viewport to collapse.

### Fix Summary  
The fix applies height for virtual scroll in two places to ensure a valid sizing chain:
1. Table container inline style in tablestyle.ts now sets height when virtualScroll is enabled and scrollHeight is not flex.
2. Table host binding in table.ts now applies the same conditional height to the host element.
3. Regression test added in table.spec.ts for virtual scroll with scrollHeight set to 100%.

### Reproduction
Change `[virtualScroll]="true"` to `false` and it does render:
https://stackblitz.com/edit/github-qb3dzpr6-92f9fuq6?file=src%2Fapp%2Fapp.component.html

### Testing  
Validated by adding a table unit test for virtualScroll with 100% scrollHeight and verifying the virtual scroller viewport remains visible.  
Manually verified against the provided StackBlitz reproduction behavior.

### Affected Version  
PrimeNG v21

### Checklist  
- [x] Defect fix only (no feature scope)  
- [x] Linked to existing issue #79  
- [x] Added regression test for the reported scenario  
- [x] Verified behavior against provided reproduction

> Migrated from `primefaces/primeng#19536` — originally by `@karoc123` on 2026-04-20

---
*Original base: `master` @ `451ca15`, head: `issue-19485` @ `d647da4`*

### DIFF
--- a/packages/primeng/src/table/style/tablestyle.ts
+++ b/packages/primeng/src/table/style/tablestyle.ts
@@ -231,7 +231,8 @@ const classes = {
 
 const inlineStyles = {
     tableContainer: ({ instance }) => ({
-        'max-height': instance.virtualScroll ? '' : instance.scrollHeight,
+        'max-height': instance.virtualScroll ? null : (instance.scrollHeight ?? null),
+        height: instance.virtualScrollViewportHeight,
         overflow: 'auto'
     }),
     thead: { position: 'sticky' },
--- a/packages/primeng/src/table/table.spec.ts
+++ b/packages/primeng/src/table/table.spec.ts
@@ -193,6 +193,91 @@ describe('Table', () => {
 
     @Component({
         standalone: false,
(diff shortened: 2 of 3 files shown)


karoc123:
I'm the author of the PR on the old repository: https://github.com/primefaces/primeng/pull/19536
I still would like to fix this, feel free to discuss the work with me. :)

Created new clean PR here: https://github.com/openng-foundation/open-prime/pull/1232

geromegrignon:
thanks @karoc123, closing this one :)

PR #1003, opened 2026-06-30
fix: #19495 - listbox content overflows container when filter or header is present

geromegrignon:
This PR fixes the listbox content overflow issue that occurs when a filter or header is present alongside scrollHeight. The root .p-listbox element now uses a flex column layout so the list container properly fills the remaining space after the header/filter instead of overflowing.

Fixes [#72](https://github.com/primefaces/primeng/issues/72)

> Migrated from `primefaces/primeng#19529` — originally by `@alex1972000` on 2026-04-16

---
*Original base: `master` @ `8fb53b4`, head: `fix/listbox-scroll-overflow` @ `a47fffa`*


geromegrignon:
Closed as too old.

PR #1006, opened 2026-06-30
Fix tooltip with template lifecycle

geromegrignon:
### Summary
This PR fixes two issues in the `Tooltip` directive:

1. Template-based tooltip content was rendered through the host `ViewContainerRef`, which could leave embedded views alive after the tooltip DOM was removed.
2. `unbindContainerMouseleaveListener()` was rebinding the listener instead of calling the stored unlistener.

### Problem
For `TemplateRef` tooltips, `updateText()` used `ViewContainerRef.createEmbeddedView()`, then moved the rendered DOM nodes into the tooltip overlay container. That created a lifecycle mismatch:
- the embedded view was attached to the host view container
- the tooltip overlay manually moved its DOM nodes elsewhere
- the created view was never stored or destroyed

This could lead to stale template content accumulating or ending up back in unrelated DOM during churn.

Separately, `unbindContainerMouseleaveListener()` contained a clear bug:
```ts
if (this.containerMouseleaveListener) {
    this.bindContainerMouseleaveListener();
    this.containerMouseleaveListener = null;
}
```

### Fix
This PR changes tooltip template rendering to let the tooltip own the embedded view lifecycle:

- use `TemplateRef.createEmbeddedView()` instead of `ViewContainerRef.createEmbeddedView()`
- store the created `EmbeddedViewRef`
- clear previous tooltip content before every re-render
- destroy the embedded view during tooltip cleanup
- null out `tooltipText` during removal
- fix `unbindContainerMouseleaveListener()` to call the stored unlistener

### Why this is better
- avoids inserting tooltip template views into the host container
- keeps overlay content lifecycle local to the tooltip
- prevents duplicate template DOM on repeated updates
- properly removes `mouseleave` listeners for `autoHide=false`

### Tests
Added regression coverage for:
- template tooltip content does not duplicate when `updateText()` runs multiple times
- container `mouseleave` listener is removed during cleanup

### Verification
I verified the new tests fail against the old implementation and pass with this fix.

Results:
- before fix: 2 failures in the new regression tests
- after fix: `46 SUCCESS`

> Migrated from `primefaces/primeng#19515` — originally by `@ericleib` on 2026-04-02

---
*Original base: `master` @ `85f00f7`, head: `fix/tooltip-template-lifecycle` @ `84c3775`*

### DIFF
--- a/packages/primeng/src/tooltip/tooltip.spec.ts
+++ b/packages/primeng/src/tooltip/tooltip.spec.ts
@@ -514,6 +514,18 @@ describe('Tooltip', () => {
         it('should handle template content', () => {
             expect(tooltipDirective.content).toEqual(component.tooltipTemplate);
         });
+
+        it('should not duplicate template content when rerendering', () => {
+            tooltipDirective.tooltipText = document.createElement('div');
+            tooltipDirective.setOption({ tooltipLabel: component.tooltipTemplate });
+
+            tooltipDirective.updateText();
+            expect(tooltipDirective.tooltipText.querySelectorAll('.custom-tooltip').length).toBe(1);
+
+            tooltipDirective.updateText();
+
+            expect(tooltipDirective.tooltipText.querySelectorAll('.custom-tooltip').length).toBe(1);
+        });
     });
 
     describe('Tooltip Options', () => {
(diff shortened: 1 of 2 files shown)


geromegrignon:
Closed as too old.

PR #1017, opened 2026-06-30
fix: recursive sort-call for treetable in multisort-mode with custom sort function

geromegrignon:
I've noticed that the custom-sort-function in treetable is supposed to be called for each node and children, but because of a wrong variable reference the sort function is called multiple times for the same input. This PR fixes that.

> Migrated from `primefaces/primeng#19471` — originally by `@ADegele` on 2026-03-10

---
*Original base: `master` @ `f06887c`, head: `treetable-custom-multisort` @ `3d47ee8`*


geromegrignon:
Closed as too old.

PR #1022, opened 2026-06-30
fix(19438): forceSelection should work even when input field gets emptied

geromegrignon:
Fixes [#19438 ](https://github.com/primefaces/primeng/issues/19438)

Currently, when the input field was dispatching a `change` event, the bound method `updateInputWithForceSelection` would return prematuraly if the field is empty.this is a problem as it will prevent a freshly emptied field to clear the inner autocomplete modelValue. Also, clearing the input field should not be bound to the minQueryLength condition.

The fix is focused on 2 cases : an input being cleared (value != defaultValue) or having a given value. 

I also added a case within the tests that verifies that a multiple & forceSelection autocomplete will empty the inputField value but not the autocomplete modelValue if such situation happens.

> Migrated from `primefaces/primeng#19440` — originally by `@Cr3aHal0` on 2026-02-26

---
*Original base: `master` @ `f11b0ce`, head: `fix-19438-autocomplete-clear-when-force-selection` @ `c8debc9`*

### DIFF
--- a/packages/primeng/src/autocomplete/autocomplete.spec.ts
+++ b/packages/primeng/src/autocomplete/autocomplete.spec.ts
@@ -1529,6 +1529,7 @@ describe('AutoComplete', () => {
             await testFixture.whenStable();
 
             const autocompleteInstance = testFixture.debugElement.query(By.directive(AutoComplete)).componentInstance;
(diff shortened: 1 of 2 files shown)


geromegrignon:
Closed as too old.

PR #1024, opened 2026-06-30
Refactor datepicker to also use ng-container for previousIconTemplate

geromegrignon:
**Refactor datepicker to also use ng-container for previousIconTemplate in stead of span**

When providing custom icons for the datepicker component, you can do so by providing a previous and a next icon template (amongst others).
I noticed styling was off between both my icons, so I checked the sourcecode.
It turns out the "previous" icon template gets wrapped in a span but the next icon is wrapped in an ng-container tag.

This extra span made it so my icon is slightly higher up than where it needs to be (compared to the button it gets put in).

> Migrated from `primefaces/primeng#19430` — originally by `@TimGeerts` on 2026-02-24

---
*Original base: `master` @ `f11b0ce`, head: `feature/datepicker-previousicon-template` @ `4ca89b3`*


geromegrignon:
Closed as too old.

PR #1029, opened 2026-06-30
fix(progressbar): use dynamic transition for fast value updates

geromegrignon:
The root cause is in the CSS:

```css
.p-progressbar-determinate .p-progressbar-value {                                                                     
      transition: width 1s ease-in-out;                                                                                 
}
```

When the value increments faster than 1 second, each Angular binding update starts a new CSS transition from the current visual position (not the previous data value). So the bar perpetually lags — at 50%, it may visually only be at e.g. 20% because each 1s animation was interrupted mid-way. 

To fix the issue of fast updates, I implemented a simple 1-second timeout strategy. If the previous value update occurred less than a second ago, the transition property is temporarily overridden to a 100 ms ease-in-out animation. Otherwise, it uses the standard CSS transition defined above. I have written some tests that should also help to understand the written code.

closes primefaces#19224

> Migrated from `primefaces/primeng#19419` — originally by `@Will-Smith-007` on 2026-02-21

---
*Original base: `master` @ `f11b0ce`, head: `bugfix/fast-progressbar-value-updates` @ `e5411a0`*

### DIFF
--- a/packages/primeng/src/progressbar/progressbar.spec.ts
+++ b/packages/primeng/src/progressbar/progressbar.spec.ts
@@ -181,6 +181,50 @@ describe('ProgressBar', () => {
             await fixture.whenStable();
             expect(progressBarInstance.valueStyleClass).toBe('value-test-class');
         });
+
+        describe('value transition behavior', () => {
+            // beforeEach binds value=50, which starts the 1s debounce timer.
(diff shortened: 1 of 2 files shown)


geromegrignon:
Closed as too old.

PR #1031, opened 2026-06-30
fix(15240): properly close mobile menubar when clicking outside

geromegrignon:
Fixes [#15240](https://github.com/primefaces/primeng/issues/15240)

Up until then, menubar had quite some weird behaviour when on mobile mode, from v17 to v21. The root menu overlay didn't really disappear when a click outside its scope occured, sometimes going off on further hover, sometimes not.

The core of the problem was that the initial clicked was also triggering the freshly created clickOutside listener, instantaneously killing the listener +  letting the CSS class responsible for hiding the menu. The fix was to prevent that initial click (from the toggle) to bubble up to the document listener. Also, mobileActive has been modernized with signal to ease change detection.

I also added a test to effectively check if the said root overlay properly goes "unvisible" after clicking outside the menubar.

> Migrated from `primefaces/primeng#19416` — originally by `@Cr3aHal0` on 2026-02-20

---
*Original base: `master` @ `f11b0ce`, head: `fix-15240-mobile-menubar-wont-close-on-click-outside` @ `9eec636`*

### DIFF
--- a/packages/primeng/src/menubar/menubar.spec.ts
+++ b/packages/primeng/src/menubar/menubar.spec.ts
@@ -467,18 +467,18 @@ describe('Menubar', () => {
             expect(menuButton).toBeTruthy();
         });
 
-        it('should toggle mobile menu on button click', () => {
+        it('should toggle mobile menu on button click', async () => {
             menubarInstance.queryMatches.set(true);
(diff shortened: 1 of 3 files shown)


geromegrignon:
Closed as too old.

PR #1032, opened 2026-06-30
fix(19405): messageService.clearAll() should properly remove toasts with lifetime

geromegrignon:
Fixes [19405](https://github.com/primefaces/primeng/issues/116)

Previously, calling `clearAll()` would not follow the same principles as calling the manual close button (via `onCloseIconClick`), making the toast flash (most likely because the timeout is never cleared, thus the motion never emits the leave event, leading into the parent toast not removing the toastitem instance despite the toastitem `visible()` being set to false)

> Migrated from `primefaces/primeng#19413` — originally by `@Cr3aHal0` on 2026-02-19

---
*Original base: `master` @ `f11b0ce`, head: `fix-toast-clearall-from-outside-closefn` @ `a5df530`*


geromegrignon:
Closed as too old.

PR #1225, opened 2026-07-01
chore: fix build error for overlay component

gianluca-moro:
Fixes build error for overlay component.

All build scripts in the package.json now run successfully.


geromegrignon:
Thanks!

PR #1230, opened 2026-07-02
feat: Angular 22 signal forms support, component modernization, ESLint 10 & tooling

### LINKED ISSUE #555: VirtualScroll Support for PickList
### Describe the bug

Picklist doesn't allow usage of listbox virtual scroll

### Pull Request Link

[_No response_](https://github.com/primefaces/primeng/pull/1177)

### Angular version

20

### PrimeNG version

v19

### Node version

_No response_

### Browser(s)

_No response_

Check picklist properties

### Expected behavior

Picklist properties allow configuration for virtual scroll

> Migrated from `primefaces/primeng#17873` — originally by `@DaSchTour` on 2025-03-13

### LINKED ISSUE #558: leadingZero attribute for InputNumber
It would be very helpful for us, if InputNumber could support the option to leave leading zeros alone, instead of truncating them automatically. Would this be possible to implement? I could not find any way of currently using primeng for this.

If the field contains the number: `305,00`, and the `3` is deleted, primeng automatically formats the number to `5,00`.
What I would like is that the remaining number stays `05,00` instead of `5,00`.  (i.e. the leading 0's should **not** be truncated by primeng)

This is how number-inputs naturally behave on the web, and therefore currently confusing for people who work daily with other fields using their keyboard.

> Migrated from `primefaces/primeng#17823` — originally by `@jasper-wd` on 2025-03-05

### LINKED ISSUE #559: Size support for autocomplete in multiple mode
### Describe the bug

When using p-autocomplete with [multiple]="true", the size="large" property doesn't have any effect. This seems to be a bug where size attributes aren't properly applied to multi-select autocomplete components.

**Doesn't works:**
`<p-autocomplete
  inputId="groups"
  [suggestions]="suggestedGroups()"
  formControlName="groups"
  [showClear]="true"
  appendTo="body"
  [multiple]="true"
  (completeMethod)="onCompleteGroups($event)"
  [completeOnFocus]="true"
  styleClass="w-full"
  size="large">
</p-autocomplete>`

**Works:**
`<p-autocomplete
  inputId="groups"
  [suggestions]="suggestedGroups()"
  formControlName="groups"
  [showClear]="true"
  appendTo="body"
  (completeMethod)="onCompleteGroups($event)"
  [completeOnFocus]="true"
  styleClass="w-full"
  size="large">
</p-autocomplete>`

### Pull Request Link

_No response_

- [x] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/github-watunekb?file=src%2Fapp%2Fapp.component.html

### Environment

**Node:** 20.17.0
**PrimeNG Version:** 19.0.6
**Angular Version:** 19.1.5
**Browser:** Google Chrome
**Device:** MSI Katana GF66 12UE

### Angular version

19.1.5

### PrimeNG version

v19

### Node version

20.17.0

### Browser(s)

Chrome

  1. Create a p-autocomplete component with multiple=true
  2. Set size="large"
  3. Observe that the size attribute doesn't affect the component styling

### Expected behavior

The autocomplete component should respect the size="large" property and render with larger input field height, text, and padding, similar to how it works when multiple=false.

> Migrated from `primefaces/primeng#17818` — originally by `@JuamBer` on 2025-03-05

### LINKED ISSUE #569: Improve button directive reactivity
### Describe the bug

When i change the outlined value by a variable, the ngOnChanges performs but without changing classes. It works only on init.

### Pull Request Link

_No response_

- [x] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/u8eg5e42-5wnueawc

### Environment

Windows 24h2

### Angular version

19

### PrimeNG version

v19

### Node version

_No response_

### Browser(s)

_No response_

1. Make a button with pButton
2. Set outlined based on variable or any dynamic behaviour
3. Change the value

### Expected behavior

Should dynamically update classes

> Migrated from `primefaces/primeng#17785` — originally by `@benoitjry` on 2025-02-27

### LINKED ISSUE #583: Menubar - keyboard navigation enter doesn't work
### Describe the bug

There is an issue with keyboard navigation in menubar when using custom template for menu item. We are unable to trigger menu command or route navigation using enter or space key. Event is blocked somewhere even if we add (keydown.enter) handler inside custom template.

### Pull Request Link

_No response_

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/hcdxezjf?file=src%2Fapp%2Fimports.ts,src%2Fapp%2Fmenubar-template-demo.ts,src%2Fapp%2Fmenubar-template-demo.html

### Environment

Win 11

### Angular version

19.0.1

### PrimeNG version

v19

### Node version

_No response_

### Browser(s)

_No response_

1. Go to https://stackblitz.com/edit/hcdxezjf?file=src%2Fapp%2Fimports.ts,src%2Fapp%2Fmenubar-template-demo.ts,src%2Fapp%2Fmenubar-template-demo.html
2. Focus on menubar.
3. Naviagate throught menu using left arrow, right arrow, down arrow.
4. Try to trigger command on any menu item using enter. 

You should see console log with label in console but nothing appears. But if you click on menu item with mouse, console log is triggered.

### Expected behavior

User should be able to trigger enter event or command of MenuItem using only keyboard.

> Migrated from `primefaces/primeng#17672` — originally by `@GitJacek` on 2025-02-13

### LINKED ISSUE #591: ConfirmationService: Showing dialog immediately after another was closed
### Describe the bug

This issue was already observed here: https://github.com/primefaces/primeng/issues/15916 and also occurs in the latest version.

"Calling confirmationService.confirm(...) twice in quick succession will cause the second dialog to not be opened at all or immediately close itself. Only a generous setTimeout of 100ms to 500ms depending on browser speed will fix this."

### Pull Request Link

_No response_

- [x] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/xcaun3-phdkjukd?file=src%2Fapp%2Fconfirm-dialog-basic-demo.ts

### Environment

- 

### Angular version

19.1.5

### PrimeNG version

v19

### Node version

18.20.3

### Browser(s)

_No response_

 1. Click on button "Does not work". Confirm "Yes" => Second dialog does not show.
2. Click on button "Works". Confirm "Yes" => Second dialog shows after timeout.

(Demo was forked from https://github.com/primefaces/primeng/issues/15916 and versions were upgraded)

### Expected behavior

Second dialog should show right after closing the first without the need to specify a timeout.

> Migrated from `primefaces/primeng#17651` — originally by `@f-rumpold` on 2025-02-12

### LINKED ISSUE #592: Radio button not centered on primeng.org
### Describe the bug

This is regarding an issue I believe was wrongly closed: https://github.com/primefaces/primeng/issues/7873

Me and others in my organization all have this same issue where the radio button inner circle isn't centered. Attached is a video of the bug in action, this becomes especially noticable when zooming in or out. I was able to recreate this also in the attached Stackblitz repos.

https://github.com/user-attachments/assets/74c06c5b-460d-40ef-b3f6-b16952947e20

### Pull Request Link

_No response_

- [ ] Lack of time
- [x] Other

### Other Reason

I am sadly not knowledgable enough to know how to correct this. I don't even know what causes this. If it helps I found a similar bug solved on a different component library: https://github.com/carbon-design-system/carbon/pull/6097/files

### Reproducer

https://stackblitz.com/edit/github-ze9i24ka-z8m17tsk?file=src%2Fapp%2Fapp.component.html

### Environment

My display resolution is 1920x1080 with a scale of 125%. Changing scale or resolution didn't do anything. However on my 4k screen the button seems to always have a centered inner circle.

### Angular version

19

### PrimeNG version

v19

### Node version

_No response_

### Browser(s)

Microsoft Edge for Business Version 132.0.2957.140, Firefox 135.0

1. Go to https://primeng.org/radiobutton or provided Stackblitz repos
2. Change preset to Material (this theme shows it the most obvious)
3. Zoom in and out (ctrl + scroll), you should see the button not be centered anymore. 

### Expected behavior

The button inner circle should stay centered.

> Migrated from `primefaces/primeng#17642` — originally by `@Wiemarp` on 2025-02-11

### LINKED ISSUE #596: p-checkbox: Style class not changing when readonly
### Describe the bug

When the checkbox is disabled, I make design changes in the style file using the '.p-checkbox .p-checkbox-box:read-only' selector. However, when readonly is false in html, it still takes the ':read-only' class and overrides my checkbox styles.

### Pull Request Link

_No response_

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/github-kewflj8f?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.component.css

### Environment

Not relevant

### Angular version

18.1.3

### PrimeNG version

v17 (LTS Only)

### Node version

23.6.1

### Browser(s)

Chrome 133

When code is written with '.p-checkbox .p-checkbox-box:read-only' in the style, this class is added first and other styles are overridden even if read-only is false.

### Expected behavior

When the checkbox is not read-only, it should not take the read-only style

> Migrated from `primefaces/primeng#17628` — originally by `@talha-al` on 2025-02-10

### LINKED ISSUE #613: PrimeNG p-multiSelect: Keep Selected Items on Top
I'm using PrimeNG's p-multiSelect component with multiple selection enabled. Every time I select an item, I don't want to search for it again in the list. Instead, I want the selected items to always appear at the top of the dropdown.

Current Behavior:

When I select multiple items, they stay in their original position in the list.
If the list is long, I have to scroll or search again to find my selected items.
Expected Behavior:

Selected items should automatically move to the top of the list.
The remaining unselected items should stay below the selected ones.
What I've Tried:
I tried sorting the options manually when the selection changes, but I couldn't achieve the desired behavior.

```
<p-multiSelect 
  [options]="sortedOptions" 
  [(ngModel)]="selectedValues" 
  optionLabel="name" 
  (onChange)="sortSelectedOptions()">
</p-multiSelect>



sortedOptions: any[] = [];
selectedValues: any[] = [];

options = [
  { name: 'Apple', value: 'apple' },
  { name: 'Orange', value: 'orange' },
  { name: 'Banana', value: 'banana' },
  { name: 'Mango', value: 'mango' }
];

ngOnInit() {
  this.sortedOptions = [...this.options];
}

sortSelectedOptions() {
  this.sortedOptions = [
    ...this.selectedValues, 
    ...this.options.filter(opt => !this.selectedValues.includes(opt))
  ];
}


```

> Migrated from `primefaces/primeng#17544` — originally by `@CavidH` on 2025-01-31

### LINKED ISSUE #623: p-table with selectionMode="single" gives hover even with [rowHover]="false"
### Describe the bug

Expected: Row hover should disable with [rowHover]="false" even with selectionMode="single".
(The same holds true for selectionMode="multiple")

p-table:
https://primeng.org/table#single-selection

stackbliz example:
https://stackblitz.com/edit/swl9a6qv?file=src%2Fapp%2Ftable-single-selection-demo.html

Version: 19 and before (i also tested 18)

### Pull Request Link

_No response_

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/swl9a6qv?file=src%2Fapp%2Ftable-single-selection-demo.html

### Environment

Stackblitz

### Angular version

19.0.0

### PrimeNG version

v19

### Node version

_No response_

### Browser(s)

Chrome 132

Hover over rows

### Expected behavior

No hover style

> Migrated from `primefaces/primeng#17468` — originally by `@Dorus` on 2025-01-22

### LINKED ISSUE #635: PanelMenu: Separator: true does not showing inside submenu items
### Describe the bug

The PanelMenu with the attribute separator set to true is only working in a primary tree level (not working inside submenu). I can't seem to find a way to display a separator between submenu items.

Image

### Pull Request Link

_No response_

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Other

### Other Reason

Problem already reported in [#15817](https://github.com/primefaces/primeng/issues/15817), but it was not carried forward.

I tested it by adding other simple components (overwriting the separator with title <h> tags, they are also not displayed). It's not something exclusive to dividers, apparently.

### Reproducer

https://stackblitz.com/edit/github-bkyvdrm5

### Environment

Stackblitz Demo or Local environment

### Angular version

19.1.1

### PrimeNG version

v19

### Node version

22.13.0

### Browser(s)

Edge 131.0.2903.146, Chrome 131.0.6778.265

[Stackblitz already configured](https://stackblitz.com/edit/github-bkyvdrm5)

### Expected behavior

The separator must also be usable in the internal levels of the panel menu (in the submenus).

> Migrated from `primefaces/primeng#17430` — originally by `@luisotavio-dev` on 2025-01-18

### LINKED ISSUE #650: [v19] Template naming unconsistent bewteen Select, MultiSelect and AutoComplete
### Describe the bug

In the v17, we could make a template for selectedItems, for example a flag as you can see in v17 the MultiSelect's demo.
In the current v19.0.5, the template dont work anymore.

The feature was usefull for my work, it would be great if can be fixed :)

### Pull Request Link

_No response_

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/u84gec32-9td8bkwa

### Environment

Windows 24H2

### Angular version

19.0.6

### PrimeNG version

v19

### Node version

20.18.1

### Browser(s)

Google Chrome

1. Make a multiselect input
2. Inside put a ng-template tag with #selectedItem
3. Make a custom template

### Expected behavior

It should show the #selectedItem template as in v17 primeng

> Migrated from `primefaces/primeng#17358` — originally by `@benoitjry` on 2025-01-12

### LINKED ISSUE #653: InputNumber Increment/Decrement Disabled Style Bug
### Describe the bug

After the update to PrimeNg v18, the Input Number component no longer shows the increment/decrement buttons as disabled, only the input.

The buttons can still be focused and clickable, even though functionality is not triggered when clicked, but they are misleading and should be styled as before with lower opacity (0.6), pointer-events: none, and cursor: default.

Primeng v17:
https://stackblitz.com/edit/dgmwhhmz?file=src%2Fapp%2Finput-number-buttons-demo.html
image

PrimeNg v18:
https://stackblitz.com/edit/qfymj4qd?file=src%2Fapp%2Finput-number-buttons-demo.html
image

### Pull Request Link

_No response_

- [X] Lack of time
- [X] Difficulty understanding the codebase
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/qfymj4qd?file=src%2Fapp%2Finput-number-buttons-demo.html

### Environment

Ubuntu 24.10

### Angular version

18.2.10

### PrimeNG version

v18 (LTS Only)

### Node version

20.18.1

### Browser(s)

Edge and Chrome 131

Set [disabled]="true" and [showButtons]="true" on the input number component, both modes (stacked, or horizontal)

### Expected behavior

Increment/decrement buttons should be disabled when the input number component has the disabled attribute set to true.

> Migrated from `primefaces/primeng#17353` — originally by `@Joebeurg` on 2025-01-10

### LINKED ISSUE #757: Component: inputNumber  - Using property 'maxLength'  prevents from using decrement and increment buttons
### Describe the bug

Using property 'maxLength'  prevents from using decrement and increment buttons of the component inputNumber

### Environment

Angular CLI: 18.2.4
Node: 18.20.2
Package Manager: npm 10.5.0
OS: win32 x64

Angular: 18.2.4
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
`@angular-devkit`/architect       0.1802.4
`@angular-devkit`/build-angular   18.2.4
`@angular-devkit`/core            18.2.4
`@angular-devkit`/schematics      18.2.4
`@schematics`/angular             18.2.4
rxjs                            7.8.1
typescript                      5.5.4
zone.js                         0.14.10

### Reproducer

_No response_

### Angular version

18.2.4

### PrimeNG version

17.18.11

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.20.2

### Browser(s)

Chrome

- Create an inputNumber add the property [showButtons] = true
- Try  using the decrement and increment buttons ( it works perfectly )
- Now Add the property [maxLength]="3" to this inputNumber
- Now click on the decrement and increment buttons ( Value is bot changing as it is suppose to)

### Expected behavior

The value of the inputNumber should change by clicking on the increment and decrement buttons even when it has the following properties :  [showButtons]="true" and [maxLength]="3"

`    <p-inputNumber
        form-control
        formControlName="xxx"
        [showButtons]="true"
        [maxLength]="3" />`

> Migrated from `primefaces/primeng#16715` — originally by `@Nicolasprrt` on 2024-11-08

### LINKED ISSUE #869: Component: multiSelect
### Describe the bug

In the new version of the multiSelect component, when a selected value does not match any of the values in the options, it displays as a 'null' label, whereas in previous versions nothing was displayed, making the view cleaner. Is there a way to prevent the label from appearing in the newer versions as well?
bug-primeNg-multiSelect

### Environment

We're using primeNg v17.6.0, integrated with Formly v6.3.0
Environment

### Reproducer

https://stackblitz.com/run?file=src%2Fapp%2Fmulti-select-reactive-forms-demo.ts,src%2Fapp%2Fmulti-select-reactive-forms-demo.html%3AL9

### Angular version

17.1.0

### PrimeNG version

17.6.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.13.0

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

> Migrated from `primefaces/primeng#16288` — originally by `@DemSantoro` on 2024-08-26

### LINKED ISSUE #935: Tree table selection is not working for Shift key
### Describe the feature you would like to see added

Could you please take a look into it.
Multiple selection in Tree table with metaKeySelection = "true" is not working for Shift key.
https://primeng.org/treetable#selectionmultiple

### Is your feature request related to a problem?

_No response_

### Describe the solution you'd like

_No response_

### Describe alternatives you have considered

_No response_

### Additional context

_No response_

> Migrated from `primefaces/primeng#13690` — originally by `@lk-granit` on 2023-09-18

### LINKED ISSUE #936: MultiSelect Checkbox
### Describe the bug

When checkbox is checked and click he close popup.

i added
```
document.addEventListener('click', event => {
      if (event.target instanceof HTMLElement && event.target.classList.contains('p-checkbox-box')) {
        event.stopPropagation();
      }
    });
```
and now is working fine

### Environment

Local Setup, Mac OS
Angular CLI: 16

### Reproducer

_No response_

### Angular version

16.0.4

### PrimeNG version

16.0.0

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.12.1

### Browser(s)

Brave

_No response_

### Expected behavior

_No response_

> Migrated from `primefaces/primeng#13216` — originally by `@jhcpeixoto` on 2023-06-16

### LINKED ISSUE #938: Autocomplete: selectedItem template only works if multiple="true"
### Describe the bug

According to the documentation, there should be a `selectedItem`-Template, that is used for displaying the selected object from an autocomplete input. Currently this template is ignored and instead the [field]-attribute is used. This does not allow to specify a complex template when the `suggestions` are complex objects.

This issue has also been reported in #2242 
`@mertsincan` said a new ticket should be opened if the issue still persists.

### Environment

Current version of angular and primeng, happening in all browsers.

### Reproducer

https://github-cnxtrn.stackblitz.io

### Angular version

15.2.8

### PrimeNG version

15.4.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

18.16.0

### Browser(s)

_No response_

1. Open Stackblitz or use this snippet in your application:

```html
<p-autoComplete
  [(ngModel)]="customer"
  (completeMethod)="searchCustomers($event)"
  [suggestions]="customerSuggestions"
  [forceSelection]="true"
  field="last_name"
>
  <ng-template pTemplate="selectedItem" let-customer>
    <div>{{ customer.first_name }} {{ customer.last_name }}</div>
  </ng-template>

  <ng-template let-customer pTemplate="item">
    <div>{{ customer.first_name }} {{ customer.last_name }}</div>
  </ng-template>
</p-autoComplete>
```

2. See the input value of a selected item being "Doe" (from 'field') instead of "Jane Doe" (from 'selectedItem')

### Expected behavior

When the selectedItem is specified it should be used instead of the field-attribute.

> Migrated from `primefaces/primeng#12983` — originally by `@Kev1n337` on 2023-05-03

### LINKED ISSUE #1228: Toast only works for 'stack' mode but not 'single'
The toast does not show up if you use the default mode `single`, it's only visible if you use `stack` mode. 

https://github.com/user-attachments/assets/89291704-390e-43c8-b9ba-b1b36b49e7dc

It used to work in v21 but with the introduction of the `mode` in v22 it broke some of the functionality.

nesgarbo:
## Overview

Brings the library up to Angular 22 with first-class **Signal Forms** support,
finishes the migration of components to modern signal APIs, refreshes the dev
tooling, and fixes/updates the MCP server and docs.

All form paradigms remain interchangeable: template-driven (`ngModel`), reactive
(`formControl`) and the new signal forms (`[formField]`) work through the same
`ControlValueAccessor` — no compatibility was removed.

## Signal Forms

- `BaseEditableHolder` now exposes `readonly`, `touched` and `errors` inputs that
  the `[formField]` directive binds automatically (alongside the existing
  `required`, `invalid`, `disabled`, `name`).
- `BaseInput`: `pattern` accepts `string | RegExp | (string|RegExp)[]` and is
  normalized to `readonly RegExp[]` to match the Angular `FormValueControl`
  contract (fixes the previous `string`-typed pattern). Adds `minLength`/`maxLength`
  (the names bound by `[formField]`) and deprecates lowercase `minlength`/`maxlength`.
- DatePicker honors the signal-forms `min`/`max` constraints as its selectable
  date range.
- New integration spec exercising `[formField]` on `pInputText`, checkbox and
  `p-inputmask`.

## Component modernization

- Removed the last decorator-based APIs: **0** remaining `@Input`/`@Output`/
  `@ViewChild`/`@ContentChild`/`@HostBinding`, and `@HostListener` migrated to
  `host: {}` metadata across 35 files.
- Constructor DI migrated to `inject()` in the remaining components/services.
- `steps`, `dragdrop`, `avatargroup`, `baseicon` and `PrimeTemplate` moved to
  `input()`/`output()`/`viewChild()`/`model()` and built-in control flow.

## Tooling

- ESLint 9 → 10 with a **flat config** (`eslint.config.mjs`), fixing the lint
  script that still used the removed `--ext`/`--ignore-path` flags — linting runs
  again for the first time under Angular 22.
- `@commitlint` 19 → 21, `cross-env` 7 → 10, `glob` 10 → 13, `lint-staged` 12 → 17,
  plus prettier/tsup/fs-extra/husky bumps.
- Dropped the unused `@angular/animations` dev dependency.

## MCP & docs

- Fixed the MCP declaration build (`ignoreDeprecations` for TS 6) and added the
  `migrate_v21_to_v22` tool + v21→v22 guide.
- Added a Signal Forms example to InputText and a new **v22 migration guide** in
  the showcase.

## Breaking changes

- **Inputs are now read-only signal inputs.** Assigning to a component/directive
  input imperatively (e.g. `iconRef.spin = true` / `iconRef.spin.set(true)`) no
  longer works — bind through the template (`[spin]="..."`). This is the
  library-wide Angular 22 direction; documented in the v22 migration guide.
- Lowercase `minlength`/`maxlength` deprecated in favor of `minLength`/`maxLength`.
- `pattern()` now returns `readonly RegExp[]` instead of a string.

## Testing

- Library build is green (all entry points).
- Repaired several pre-existing specs that prevented the unit suite from
  compiling, plus a `sourceRoot` fix so the karma builder resolves the test
  bootstrap.
- The `@HostListener` migration was verified against a pre-change baseline:
  **0 new test failures** (identical pre-existing failures before/after).

## pnpm 11

- **pnpm 9 → 11** (`packageManager: pnpm@11.10.0`, `engines: node >=22.13`, CI on
  Node 22.x). The blockers noted earlier are resolved by declaring the phantom
  dependencies pnpm 9's hoisting papered over: `chart.js`/`quill` at the root
  (imported by `primeng/chart` and `primeng/editor`), `@types/jasmine`/`jasmine-core`
  in `packages/primeng`, and `@primeicons/angular`/`@primeuix/styled`/`@primeuix/utils`
  in the showcase. Dependency build scripts blocked by pnpm 10+ are approved
  explicitly via `allowBuilds` (esbuild, @parcel/watcher, lmdb, msgpackr-extract),
  which also fixes the esbuild postinstall failure previously attributed to Node 26.

## Review follow-ups

- Signal forms example fixed per review: no manual `[invalid]` alongside
  `[formField]` (NG8022), native `(submit)` instead of `ngSubmit`, and the demo
  code block now renders (`demos.json` + signal-forms support in
  `build-demo-code.mjs`).
- Components that accepted the `readonly` state bound by `[formField]` without
  honoring it now do: DatePicker, RadioButton, ToggleButton, SelectButton,
  Slider, ColorPicker, CascadeSelect, TreeSelect.
- Migration guides corrected (the `@angular/animations` removal was a v21 change;
  `PrimeTemplate.type/name` signal reads and the `readonly()` default change are
  documented; `/migration/v22` linked in the menu).
- Flat config now lints `.js/.mjs` scripts (dead code cleaned up, latent
  `ReferenceError` in `build-helper.mjs` fixed).

## Closed issues

Backlog items from the contribution board fixed in this PR:

- Fixes #935 — TreeTable had no shift-key range selection in multiple mode; it now mirrors Table's anchor/range logic over the visible flattened rows.
- Fixes #1228 — Toast was invisible in `single` mode: the rc.1 styles assume the stacked model's `data-mounted`/`data-visible` attributes; scoped overrides restore in-flow layout and the enter/leave keyframes for non-stacked toasts.
- Fixes #635 — PanelMenu rendered separator `<li>`s inside submenus but the theme has no CSS for `.p-menuitem-separator`, so they were invisible; added the missing border rule (token-aligned with Menu's separator).
- Fixes #591 — chaining `confirm()` from another confirmation's accept callback killed the new subscription; `hide()` now only tears down the confirmation being closed.
- Fixes #938 — AutoComplete `selectedItem` template now also renders in single selection mode.
- Fixes #559 — AutoComplete `size` (sm/lg) now applies in multiple mode (root classes + container CSS, and the inner input no longer gets `size="small"` as a numeric HTML attribute).
- Fixes #869 — MultiSelect no longer renders the literal string "null" in the label when selected values have no matching option.
- Fixes #936 — MultiSelect header "select all" checkbox couldn't deselect: `selectAll` lost its `null` default in the signal migration and the strict `!== null` check never matched.
- Fixes #757 — InputNumber spin buttons compared `maxlength` against the formatted value (thousands separators, currency, decimal padding), permanently blocking spinning; the guard now uses the parsed value and always allows shrinking.
- Fixes #653 — InputNumber increment/decrement disabled styling checked the presence of `maxLength`/`minLength` instead of the value against `max`/`min`, and ignored the component's own disabled state.
- Fixes #558 — new `leadingZero` input for InputNumber (default `false`): typed or remaining leading zeros survive reformatting while editing and normalize on blur.
- Fixes #592 — RadioButton dot was off-center at fractional sizes (independent per-edge pixel rounding in the flex layout); centered via absolute positioning + transform, which renders subpixel-accurately.
- Fixes #596 — Checkbox now reflects the readonly state (`p-readonly` class + `data-p`), matching Rating's pattern.
- Fixes #555 — PickList now supports virtual scrolling (`virtualScroll`, `virtualScrollItemSize`, `virtualScrollOptions` forwarded to both inner Listboxes).
- Fixes #613 — new `selectedItemsOnTop` input for MultiSelect: selected options are pinned first when the overlay (re)opens, without reordering under the cursor.
- Fixes #650 — Select/MultiSelect/AutoComplete now accept both casings of their selected-item template name (`#selectedItem`/`#selecteditem`, `#selectedItems`/`#selecteditems`) via non-breaking aliases.
- Fixes #583 — Menubar Enter key didn't execute items using custom templates (the dispatched click never reached the `itemContent` handler); the key handler now falls back to clicking that element.
- Fixes #569 — pButton directive now reacts to dynamic changes of `outlined`/`text`/`plain`/`raised`/`rounded`/`size`/`fluid`, and removes stale variant classes when they toggle off.
- Fixes #623 — `p-table` applied row hover with `selectionMode` even when `[rowHover]="false"` (explicit `false` now wins; default behavior unchanged).


packages/primeng/src/icons/baseicon/baseicon.ts
```diff
@@ -19,13 +19,13 @@ import { BaseIconStyle } from './style/baseiconstyle';
     }
 })
 export class BaseIcon extends BaseComponent {
-    @Input({ transform: booleanAttribute }) spin: boolean = false;
```
Wiemarp, line 22: Going from the input decorator to a InputSignal could be a breaking change. This makes it so you can no longer do the following:

```ts
export class IconDirective implements OnInit {
  private readonly baseIcon = inject(BaseIcon);

  ngOnInit(): void {
    this.baseIcon.spin = true;

    // The following is not allowed on an InputSignal:
    // this.baseIcon.spin.set(true);
  }
}
```

It's bad practice to do this, but I've seen multiple projects do this.
This should probably be documented in the migration notes

nesgarbo: Thanks @Wiemarp, good catch — you're absolutely right. Moving from the `@Input` decorator to a signal `input()` is a breaking change for anyone writing the input imperatively (`this.baseIcon.spin = true`), and as you note `.set()` isn't available on an `InputSignal` either (that's only on `model()`).

I've documented this in the migration notes as you suggested: the new v22 migration guide now has a **Breaking Changes** section that calls out that inputs are read-only signal inputs and must be bound through the template (`[spin]="..."`), using your exact example (`apps/showcase/doc/migration/v22/breakingdoc.ts`). It's also mirrored in the MCP `v21 -> v22` guide. Thanks again!

nestorgarciacheste (approved):

KIRA-0 (approved):

apps/showcase/doc/inputtext/signalforms-doc.ts
```diff
@@ -0,0 +1,69 @@
...
+@Component({
+    selector: 'signal-forms-doc',
+    standalone: true,
+    imports: [FormField, InputTextModule, ButtonModule, ToastModule, MessageModule, AppCode, AppDemoWrapper],
+    template: `
+        <p-toast />
+        <app-demo-wrapper>
+            <div class="flex justify-center">
+                <form (ngSubmit)="onSubmit()" class="flex flex-col gap-4 w-full sm:w-56">
+                    <div class="flex flex-col gap-1">
```
MrZajic, line 20: I'm new in PR review for Angular. 
But for `[formField]` is not allowed `[invalid]`:
```typescript
X [ERROR] NG8022: Binding to '[invalid]' is not allowed on nodes using the '[formField]' directive [plugin angular-compiler]

    doc/inputtext/signalforms-doc.ts:29:118:
      29 │ ...d]="exampleForm.email" [invalid]="isInvalid(exampleForm.email)" />
```
Instead `[invalid]` there should be maybe:
for username:
```typescript
[class.p-invalid]="isInvalid(exampleForm.username)"
```
for email:
```typescript
[class.p-invalid]="isInvalid(exampleForm.email)"
```

apps/showcase/doc/inputtext/signalforms-doc.ts
```diff
@@ -0,0 +1,69 @@
...
+
+@Component({
+    selector: 'signal-forms-doc',
+    standalone: true,
+    imports: [FormField, InputTextModule, ButtonModule, ToastModule, MessageModule, AppCode, AppDemoWrapper],
+    template: `
+        <p-toast />
+        <app-demo-wrapper>
+            <div class="flex justify-center">
+                <form (ngSubmit)="onSubmit()" class="flex flex-col gap-4 w-full sm:w-56">
```
MrZajic: Event ngSubmit is not emitted by any applicable directives nor by `<form>` element.
Seems missing `FormsModule` in imports.

apps/showcase/doc/inputtext/signalforms-doc.ts
```diff
@@ -0,0 +1,69 @@
...
+                        @if (isInvalid(exampleForm.email)) {
+                            @for (error of exampleForm.email().errors(); track error.kind) {
+                                <p-message severity="error" size="small" variant="simple">{{ error.message }}</p-message>
+                            }
+                        }
+                    </div>
+                    <button pButton severity="secondary" type="submit"><span pButtonLabel>Submit</span></button>
+                </form>
+            </div>
+            <app-code></app-code>
```
MrZajic, line 39: `<app-code></app-code>` shows empty block in demo. Missing information in `demo.json` and maybe there is link with `build-demo-code.mjs`.

apps/showcase/doc/inputtext/signalforms-doc.ts
nesgarbo: Thanks @MrZajic, good catch! You're right that `[invalid]` can't be bound on a node that uses `[formField]` (NG8022) — the directive claims that binding for itself. Rather than switching to `[class.p-invalid]`, the manual binding turned out to be unnecessary: `[formField]` already binds the field's invalid state to `pInputText` automatically (that's exactly what the integration spec asserts in `packages/primeng/src/baseinput/signalforms.spec.ts`), so I simply removed the manual `[invalid]` bindings. Fixed in efd7454 (and the showcase now builds clean).

apps/showcase/doc/inputtext/signalforms-doc.ts
nesgarbo: Thanks, you're right — `(ngSubmit)` comes from `NgForm`/`FormGroupDirective`, and with signal forms neither is on the `<form>` element. Instead of importing `FormsModule` just for that (which would mix the template-driven paradigm into a signal forms example), I switched to the native `(submit)` event with `event.preventDefault()` in `onSubmit()`. Fixed in efd7454.

apps/showcase/doc/inputtext/signalforms-doc.ts
nesgarbo: Thanks for spotting this — you traced it exactly right. The `<app-code>` block resolves its content from `public/demos.json`, which is generated by `build-demo-code.mjs`, and that script didn't know about `@angular/forms/signals`: the `inputtext-signal-forms-demo` entry was missing entirely. I taught the generator to carry over the `@angular/forms/signals` import, add `FormField` to the component's `imports` array and handle `form(...)` property declarations, then regenerated `demos.json` — the demo now shows its code block (and the generator's test suite passes 5/5). Fixed in efd7454.

gianluca-moro (commented):
**Edit: Nevermind, this is not introduced by this PR, so I'll open a separate issue for it. You can ignore this.** 

Listbox with Checkbox has it's header checkbox misaligned

Image

gianluca-moro (approved):

packages/primeng/src/picklist/picklist.ts
gianluca-moro, line 1: There is some issue with the virtual scrolling in picklists: The scrollbar flickers and there is a blank section at the bottom of the list.

https://github.com/user-attachments/assets/1c994bd8-cc7f-4368-92b2-1ea25d69684e

nesgarbo: Thanks @gianluca-moro, great catch — both symptoms are fixed in a41ef5b.

Two things were going on:

1. **Blank gap.** The theme forces `.p-picklist .p-listbox { height: 100% }` so each list fills the flex container. With virtual scrolling the inner Listbox mounts an auto-sized `p-scroller` whose height follows its own viewport, so the 100%-tall list stretched past the scroller (as tall as the center reorder buttons) and left an empty gap under the options. I scoped a `p-picklist-virtualscroll` modifier that lets the list size to its scroller instead — verified via CDP that the list height now matches the scroller (252px vs 250px with many items; 188 vs 186 with few), with no gap.
2. **Flickering scrollbar.** That happens when `virtualScrollItemSize` isn't set: the auto-sizing scroller renders only a couple of rows and re-measures on every scroll. It's required for the scroller to size its spacer (same as Listbox). I added a **Virtual Scroll** demo to the PickList docs that sets it explicitly and documents the requirement.

Reproduced your exact scenario before/after with a headless browser; screenshots attached to the demo. Let me know if you still see anything off.

nesgarbo: Follow-up in 4c3b415 — the remaining two issues you'd still see are fixed:

- **Blank strip on the empty list & remaining gap:** the inner Listbox auto-sized its scroller, so an empty list collapsed to a thin strip. The list containers now hold a fixed height of `scrollHeight` when virtual scrolling is on, so both lists (including an empty one) stay the same height.
- **Flicker while scrolling:** PickList now forwards `autoSize: false` to the inner scrollers by default, so they no longer re-measure on every scroll. Verified via CDP that the source scroller height stays constant (250px) across scroll steps — no jitter — and both lists match. Still overridable through `virtualScrollOptions`.

gianluca-moro: @nesgarbo nice. It's almost perfect now, however I am seeing two scrollbars now

Screenshot 2026-07-06 at 08 03 13

nesgarbo: Fixed in 72444b6 — the double scrollbar is gone. The inner Listbox's list container kept its own `overflow: auto` while the virtual scroller also owns a scroll, so two scrollbars stacked up. It now sets `overflow: hidden` on the list container in virtual-scroll mode, leaving only the scroller's scrollbar. Fixed at the source in Listbox (so it also benefits a standalone virtual-scroll Listbox), reverting the earlier PickList-scoped attempt. Verified the computed container style via a unit test. Thanks for the careful review!

geromegrignon:
Hi @nesgarbo, thanks for the contribution.
It's truly appreciated, but it's impossible to properly review it as covering multiple (and way too many) topics.
We are perfectly fine with AI usage for contributions, and we encourage you to use it to generate individual PR with one focus at a time.

nesgarbo:
Thanks @geromegrignon — you're right that this covered far too many topics to review. I've split it up:

**Focused, self-contained PRs (one topic each):**
- #1253 fix(table): rowHover with selectionMode (#623)
- #1254 fix(treetable): shift-key range selection (#935)
- #1255 fix(toast): single mode (#1228)
- #1256 fix(panelmenu): submenu separators (#635)
- #1257 fix(confirmdialog): chained confirmations (#591)
- #1258 fix(radiobutton): icon centering (#592)
- #1259 fix(button): dynamic variant reactivity (#569)
- #1260 fix(multiselect): #869, #936, #613
- #1261 feat(picklist): virtual scrolling (#555)
- #1262 build(showcase): modern @angular/build dev-server
- #1263 feat(dynamicdialog): component bindings

**One cohesive PR** for the interdependent foundation (Signal Forms + component modernization + Vitest migration + pnpm 11 + ESLint 10 + v22 docs), which can't be split without breaking compilation (the migrated specs reference the new APIs, and the tooling shares package.json): #1264

Closing this one in favor of those. Happy to adjust the grouping if any of them is still too broad.

PR #1240, opened 2026-07-02
fix(scroller): blank rows when lazy loading while scrolling backward or jumping multiple pages

rgroenewoudt:
### Defect

When `p-scroller` (and components built on it, e.g. `p-table` with `virtualScrollerOptions`) is used with `lazy` + `step`, `onScrollChange` always requests exactly one page:

```ts
first: page * step
last:  (page + 1) * step
```

The scroller, however, renders `d_numToleratedItems` rows *before* the first visible index. Near a page boundary those rows belong to the **previous** page. Scrolling forward one page at a time masks this (the previous page happens to have just been loaded), but:

1. **Scrolling backward** — the page before the landing page is never requested, so the tolerated rows above the viewport are blank, and they scroll into view empty.
2. **Dragging the scrollbar across multiple pages** — only the landing page is requested; the skipped range that is partially visible stays blank.

Additionally, the emit condition `lazyLoadState.first !== newState.first || lazyLoadState.last !== newState.last` re-fires `onLazyLoad` for windows that are strict sub-ranges of the previously requested one, causing redundant duplicate fetches.

### Reproduction

- `p-scroller` or `p-table` with `[lazy]="true"`, a `step` (e.g. 100), and an `onLazyLoad` handler that fills the corresponding slice of the items array.
- Scroll down a few pages, then scroll back up: rows go blank and no `onLazyLoad` is emitted for them.
- Or grab the scrollbar and jump 3+ pages down: rows just above the landing position stay blank.

### Fix

In `onScrollChange`:

- Compute the page transition **before** `this.first` is overwritten. If the new page is anything other than the same page or exactly one page forward, include the previous page in the request: `first = (page - 1) * step`, clamped to 0.
- Only emit `onLazyLoad` when the new window actually **extends** the previously requested one (`first <` or `last >`), instead of on any difference. This prevents the widened request from being followed by a redundant narrower duplicate.

The initial-load path in `calculateNumItems` is unchanged and still seeds `lazyLoadState` with a numeric window before any scroll-driven emission, so the relational comparison never sees the initial `{}`.

### Tests
Added `Lazy Load Window Tests` to `scroller.spec.ts` covering: backward scroll widening, multi-page jump widening, unchanged sequential-forward behaviour, suppression of already-covered windows, and clamping at the first page. The first two and last two fail against the previous implementation.
These tests are generated by Claude Fable 5.

### Notes
- We have been running this fix in production as a patch since November 2025.

### DIFF
--- a/packages/optimus-ui/src/scroller/scroller.test.ts
+++ b/packages/optimus-ui/src/scroller/scroller.test.ts
@@ -1134,6 +1134,101 @@ describe('Scroller', () => {
         });
     });
 
+    describe('Lazy Load Window Tests', () => {
+        let component: TestBasicScrollerComponent;
+        let fixture: ComponentFixture<TestBasicScrollerComponent>;
+        let scroller: Scroller;
+
+        beforeEach(async () => {
+            await TestBed.configureTestingModule({
+                imports: [Scroller],
+                providers: [provideZonelessChangeDetection()],
+                declarations: [TestBasicScrollerComponent]
+            }).compileComponents();
+
+            fixture = TestBed.createComponent(TestBasicScrollerComponent);
+            component = fixture.componentInstance;
+            scroller = fixture.debugElement.query(By.directive(Scroller)).componentInstance;
+
+            component.items = Array.from({ length: 100 }, (_, i) => ({ label: `Item ${i}`, value: `item${i}` }));
+            component.lazy = true;
+            component.step = 10;
+            component.itemSize = 50;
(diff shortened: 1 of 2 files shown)


geromegrignon:
Hi @rgroenewoudt , we released 2.0 so i'll focus on reviewing PRs now.
I need another change from you, we changed the head branch from master to main, and I would need you to update the PR.

geromegrignon:
Thanks for your contribution @rgroenewoudt !

PR #1246, opened 2026-07-04
refactor: add constants to rename the project name/link

geromegrignon:
## Description

Add constants and use them for the new project name/links


apps/showcase/components/layout/topbar/app.topbar.component.ts
```diff
@@ -27,10 +28,10 @@ export class AppTopBarComponent {
 
     readonly versions = Versions;
     readonly socialLinks = [
-        { href: 'https://github.com/openng-foundation/open-prime', icon: 'pi-github' },
+        { href: GITHUB_REPO_URL, icon: 'pi-github' },
         // TODO: Invite to angular discord open-prime channel
```
dominicbachmann: We can now remove the TODO

dominicbachmann (changes requested):
nice change! I think I would have split this into separate PR's for every variable to keep the change sets smaller (something to think about in the future)

PR #1248, opened 2026-07-04
fix(autocomplete): correctly handle label in case of object and not inputLabel

### LINKED ISSUE #901: Autocomplete: "[object Object]" displayed when optionLabel input is used and the computed label value is null or undefined
### Describe the bug

The problem is at this line of code : https://github.com/primefaces/primeng/blob/e142de683b9cbcf2342774e9390ee09b0934580f/src/app/components/autocomplete/autocomplete.ts#L787C17-L787C59

It should not return the full object when the computed label is null or undefined.

### Environment

not relevant

### Reproducer

_No response_

### Angular version

18

### PrimeNG version

17.18.4

### Build / Runtime

Angular CLI App

### Language

TypeScript

latest

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

> Migrated from `primefaces/primeng#16040` — originally by `@fredpeaks` on 2024-07-17

MRinaldi9:
Close #901 and added a "one time" warning, in dev mode, in case of value of type object and inputLabel or inputValue are falsy

### DIFF
--- a/packages/optimus-ui/src/autocomplete/autocomplete.ts
+++ b/packages/optimus-ui/src/autocomplete/autocomplete.ts
@@ -6,9 +6,9 @@ import {
     computed,
     ContentChild,
     ContentChildren,
-    EmbeddedViewRef,
     effect,
     ElementRef,
+    EmbeddedViewRef,
     EventEmitter,
     forwardRef,
     HostListener,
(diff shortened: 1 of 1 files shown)


geromegrignon:
Hi @MRinaldi9 , we released 2.0 so i'll focus on reviewing PRs now.
I need another change from you, we changed the head branch from master to main, and I would need you to update the PR.

MRinaldi9:
@geromegrignon PR updated 🫡

PR #1279, opened 2026-07-07
feat(mcp): add the v21 to v22 migration guide

nesgarbo:
## Change

MCP package only:

- New `v21_to_v22` migration entry: breaking changes (Angular 22 requirement, signal-input imperative-write pitfall — `iconRef.spin = true` no longer works, `PrimeTemplate.type` is now a signal —, `pattern` normalization to `readonly RegExp[]`, `readonly` default change), deprecations (`minlength`/`maxlength` → `minLength`/`maxLength`) and what's new (Signal Forms integration).
- Advertised compatibility bumped to Angular 22+ and the migration tool description now lists v22 (docs pages for the MCP tools updated accordingly).
- `ignoreDeprecations` set in the package tsconfig so the declaration build works under TypeScript 6.

Verified: `pnpm run build` of `@primeng/mcp` passes (tsup + DTS).

> The license field of this package was intentionally left untouched.

### DIFF
--- a/apps/showcase/doc/mcp/exampleprompts-doc.ts
+++ b/apps/showcase/doc/mcp/exampleprompts-doc.ts
@@ -21,6 +21,6 @@ export class ExamplePromptsDoc {
 "How do I customize the Dialog component styling with Pass Through?"
 "Compare the Select and Listbox components"
 "What's the best component for a date picker?"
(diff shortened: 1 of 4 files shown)


gianluca-moro:
Why are you opening a second PR for the 22 migration guide if https://github.com/openng-org/open-prime/pull/1269 already exists??

PR #1286, opened 2026-07-09
refactor: apply Angular 22 update migrations

kheos31:
@geromegrignon This PR was notably submitted to prevent errors during integration into an Angular 22 application, due breaking change affecting the default behavior of component change detection.


dominicbachmann:
@kheos31 This PR is too large to be properly reviewed. Please split this into multiple smaller PRs. Although we have to wait with this in general as we are currently working on the `main` branch because the commits for Angular 22 merged onto master were causing a lot of test to fail. Thank you for your effort and engagement.

PR #1317, opened 2026-07-14
fix(showcase): correct typo json, fix error when run showcase

impossibledev2012:
- correct typo json in data/menu.json
 - fix error when run showcase
 - remove unused import component (warning when running)


geromegrignon:
Thanks for your contribution!

PR #1322, opened 2026-07-14
fix: resolve tsup entrypoint paths on windows

### LINKED ISSUE #1321: Build fails on Windows due to backslash paths in tsup configs
### Description
     When trying to build the library on Windows using `pnpm run build:lib`, the compilation fails with:
        `error TS2307: Cannot find module '@openng/optimus-ui-utils' or its corresponding type declarations.`

 ### Cause
    In `tsup.config.ts` files, `globSync` returns paths with backslashes (`\`) on Windows (e.g., `src\dom\index.ts`). The configurations use `replace(/^src\//, '')` which expects forward slashes,
      leaving the `src\` prefix intact. This causes `tsup` to build entry points inside `dist/src/` rather than `dist/`.

dewinsonbusiness:
### Description
 On Windows environments, the file paths returned by `globSync` contain backslashes (`\`). The previous regex replacing target was `/^src\//` which specifically looked for forward slashes (`/`). Because
  of this mismatch, the `src\` prefix was not stripped on Windows, causing `tsup` to build entry points inside `dist/src/` rather than `dist/`.

    This PR normalizes backslashes to forward slashes in all `tsup.config.ts` entry resolution loops, resolving build path issues on Windows and allowing the package `@openng/optimus-ui` to compile
  successfully out of the box.

Fixes #1321

  ### Key Changes
  Normalized paths using `.replace(/\\/g, '/')` before processing entry point names in:
    - `packages/optimus-ui-utils/tsup.config.ts`
    - `packages/optimus-ui-styles/tsup.config.ts`
    - `packages/optimus-ui-themes/tsup.config.ts`
    - `packages/themes/tsup.config.ts`

    ### Validation
    Verified that `@openng/optimus-ui` builds successfully on Windows with `pnpm run build:lib`.

### DIFF
--- a/packages/optimus-ui-styles/tsup.config.ts
+++ b/packages/optimus-ui-styles/tsup.config.ts
@@ -3,7 +3,7 @@ import { defineConfig } from 'tsup';
 
 const isProduction = process.env.NODE_ENV === 'production';
 
-const entry = globSync('src/**/index.ts').reduce((acc: Record<string, string>, file: string) => {
+const entry = globSync('src/**/index.ts', { posix: true }).reduce((acc: Record<string, string>, file: string) => {
     const name = file.replace(/^src\//, '').replace(/\.ts$/, '');
 
     acc[name] = file;
--- a/packages/optimus-ui-themes/tsup.config.ts
+++ b/packages/optimus-ui-themes/tsup.config.ts
@@ -4,7 +4,7 @@ import { defineConfig, type Options } from 'tsup';
 const isProduction = process.env.NODE_ENV === 'production';
 const themes: string[] = [];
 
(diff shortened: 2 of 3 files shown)


dominicbachmann (changes requested):
Thanks for pointing this out and implementing a fix! I think we can solve this more idiomatic by adding `{ posix: true }` to `globSync('src/**/index.ts'` => `globSync('src/**/index.ts', { posix: true })`. Also: please combine your commits so it is just a single one.

dominicbachmann:
@dewinsonbusiness Thank you for the fix!

dominicbachmann (approved):

dominicbachmann (changes requested):
@dewinsonbusiness Can you please fix the commitlint issue. Ready to merge after that :)

PR #1323, opened 2026-07-14
refactor: data attributes

geromegrignon:
@dominicbachmann I wonder about the exact scope and potential conflicts of refactoring primevue attributes

### DIFF
--- a/packages/optimus-ui-styled/src/utils/themeUtils.ts
+++ b/packages/optimus-ui-styled/src/utils/themeUtils.ts
@@ -182,7 +182,7 @@ export default {
                     const _css = minifyCSS((value as any).css);
                     const id = `${key}-variables`;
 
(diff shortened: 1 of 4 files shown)


dominicbachmann (commented):
@geromegrignon I assume we don't need the primevue utils anymore. I think chaning them to optimus (which was primeng) could introduce regressions?

geromegrignon:
@dominicbachmann I checked the usage: the code is meant for SSR apps to inject styles on server side.
So the renaming change is also a fix:

## Before

the server inject the CSS (primevue)
The browser checks for CSS and inject the primeng one as no 'primeng' one has been injected on server side

=> duplicate stylesheets at the end

## Now

The server inject the CSS
The browser skip injecting CSS as already present

dominicbachmann (approved):

PR #1328, opened 2026-07-14
refactor(utils): optimus renaming

geromegrignon:


dominicbachmann (approved):
I would try to keep PR's to their intended changes but okay for me. Good job!

PR #1357, opened 2026-07-17
fix: Table ColumnFilter leaves stale ZIndexUtils entries on close

EeneeS:
## Description
Fixes:
1. `ZIndexUtils.clear(this.overlay);` is not called with a null value. 
2.  `onOverlayAnimationAfterLeave` would not be called due to `this.renderOverlay.set(false);` being called to early, resulting in `ZIndexUtils.clear(this.overlay)` not being called.

## Related issues

Fixes #1356 

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that changes the public API)
- [ ] Documentation only
- [ ] Refactor, test, or chore (no user-facing change)

## Breaking changes

None

## Test plan

- [x] `npm run build`
- [ ] `npm test`
- [ ] `npm run lint`
- [x] Verified in the demo app (if applicable)

## Checklist

- [x] Issue discussed or bug clearly described (link issue when applicable)
- [ ] Tests added or updated for behavioral changes
- [ ] Documentation updated (README, JSDoc, migration notes as needed)
- [ ] Public API changes documented; breaking changes called out
- [ ] CHANGELOG updated (if the repository maintains one and the change is user-facing)
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)
- [x] I agree to follow the [OpenNG Foundation Code of Conduct](https://github.com/openng-foundation/.github/blob/main/CODE_OF_CONDUCT.md)

Issue #1356, opened 2026-07-17
[Bug]: Table ColumnFilter leaves stale ZIndexUtils entries on close


EeneeS:
### Checklist

- [x] I have searched the existing issues and have not found a duplicate
- [x] I have read the README and documentation and have not found a solution
- [x] I agree to follow the [OpenNG Foundation Code of Conduct](https://github.com/openng-foundation/.github/blob/main/CODE_OF_CONDUCT.md)

### What happened?

Table’s ColumnFilter does not properly clear its overlay from ZIndexUtils when the menu closes. In onOverlayAnimationAfterLeave, ZIndexUtils.clear(this.overlay) is called after this.onOverlayHide(). However, onOverlayHide() sets this.overlay to null, so ZIndexUtils.clear() is invoked with null instead of the actual overlay element.

As a result, the old z-index entry remains registered in ZIndexUtils. On subsequent open/close cycles, new entries are added while the stale ones are never removed, so the overlay’s assigned z-index keeps incrementing.

### Steps to reproduce

1. Open column filter and verify z-index (assume z-index = x)
2. Close column filter
3. Open column filter and verify z-index = x + 2
4. Repeat step 2-3 and verify z-index keeps incrementing.

### Expected behavior

When the ColumnFilter overlay closes, its z-index entry should be removed from ZIndexUtils, so reopening the menu does not accumulate stale entries or keep increasing the assigned z-index.

### Actual behavior

the old z-index entry remains registered in ZIndexUtils. On subsequent open/close cycles, new entries are added while the stale ones are never removed, so the overlay’s assigned z-index keeps incrementing.

### Minimal reproduction
Can be reproduced in the showcase app.

Stackblitz migrated from primeNG repo (still uses v21):
https://stackblitz.com/edit/stackblitz-starters-pztpy6q6?file=src%2Fapp%2Fapp.component.ts

### Library version

v21

### Angular version

21.0.6

### Node version

22.22.0

### Package manager

pnpm

### Browser(s)

Chrome

### Environment details

```markdown
- OS: Windows
```

### Angular application setup

None

### Angular context

- [ ] SSR (Server-Side Rendering) is involved
- [ ] Zoneless change detection is enabled

### Relevant log output

```shell

```

geromegrignon:
Thanks for reporting it!
I added it to our `catch-up` milestone as we are working on releasing v21/v22 compatible Angular versions.

PR #1376, opened 2026-07-21
 test(table): cover column filter overlay z-index cleanup on close

lukasganster:
## Description

#1357 fixed the underlying bug (`ZIndexUtils.clear(this.overlay)` ran after `onOverlayHide()` had already nulled `this.overlay`, so the stale overlay
entry was never removed and the z-index grew by 2 on every reopen), but merged without a regression test.

This PR adds that missing test. It exercises the real `ZIndexUtils` singleton and `ColumnFilter.onOverlayAnimationAfterLeave` directly: opens/closes the
overlay twice and asserts the second z-index equals the first. Against the pre-#1357 ordering this fails (second z-index is base+4 instead of base+2,
matching the reported x, x+2, x+4 growth); against current `main` it passes.

No production code changes, this only adds coverage for behavior that already shipped.

## Related issues

Relates to #1356 (fixed by #1357). This PR only backfills the regression test that PR didn't include.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation only
- [x] Refactor, test, or chore (no user-facing change)

## Breaking changes

None

## Test plan

- [x] `npm run build`
- [x] `npm test` (ran the Table suite: 6 test which fails without the fix)
- [x] `npm run lint`
- [x] Verified in the demo app (if appli

## Checklist

- [ ] Documentation updated (README, JSDoc, migration notes as needed) — not applicable
- [ ] Public API changes documented; breaking changes called out — not applicable
- [ ] CHANGELOG updated — not applicable, test-only change

## Additional context

Originally opened with a duplicate fix commit alongside this test, before #1357 merged the fix independently. Per @geromegrignon's review, rebased onto `main` and reduced to just the regression test, moved to `table.test.ts` (Karma removed).

### DIFF
--- a/packages/optimus-ui/src/table/table.test.ts
+++ b/packages/optimus-ui/src/table/table.test.ts
@@ -6,7 +6,8 @@ import { By } from '@angular/platform-browser';
 
 import { SharedModule, SortMeta } from '@openng/optimus-ui/api';
 import { Select } from '@openng/optimus-ui/select';
-import { Table, TableModule, TableService } from './table';
+import { ZIndexUtils } from '@openng/optimus-ui/utils';
+import { ColumnFilter, Table, TableModule, TableService } from './table';
 
 describe('Table', () => {
     let component: Table;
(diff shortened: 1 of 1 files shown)


jeroenpol (approved):

geromegrignon (changes requested):
Hi #1357 has been merged but the test is still welcome.

Please rebase your branch and move the test to the sibling `test.ts`file as we removed karma.

lukasganster:
Hi @geromegrignon ,
thanks for the feedback, I adapted this PR accordingly. Hope it fits now!

Regarding the Angular Signal Forms adaption mentioned in #1651 it should be fine, since its just an additional test for the bugfix.

PR #1387, opened 2026-07-23
fix(togglebutton): allow undefined as size

### LINKED ISSUE #19: ToggleButton does not allow `undefined` as size
### Describe the bug

Most components that have a `[size]` input allow it to be `undefined`, which would be the normal size.    However, the ToggleButton only allows `"small"` or `"large"` as values but not `undefined`. (https://primeng.org/togglebutton#api.togglebutton.props.size). Note that using `null` does also not work.   
This makes it hard to set it programatically. 

### Pull Request Link

https://github.com/primefaces/primeng/pull/977

### Reproducer

https://primeng.org/togglebutton#api.togglebutton.props.size

### Environment

PrimeNG 21, Angular 21

### Angular version

21.0.0

### PrimeNG version

v21

### Browser(s)

Firefox, CHrome

1. Create a `<p-togglebutton>` component and try to set the `[size]="undefined"`. It will give a build error.

### Expected behavior

ToggleButton should also allow `undefined` as size like other components (e.g., AutoComplete, Select, InputText, etc.)

> Migrated from `primefaces/primeng#19591` — originally by `@gianluca-moro` on 2026-05-21

omurcanuslu:
## Summary
- Allow `undefined` for ToggleButton `size` so the default/normal size can be set programmatically, matching other components (Checkbox, SelectButton, etc.).
- Migrate `size` to an `input()` signal and update style/tests accordingly.
- Fixes #19

## Test plan
- [ ] `[size]=undefined` / unbound size compiles without TS2322
- [ ] `size=small` / `size=large` still apply the correct CSS classes
- [ ] Existing ToggleButton unit test for size passes

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/optimus-ui/src/togglebutton/style/togglebuttonstyle.ts
+++ b/packages/optimus-ui/src/togglebutton/style/togglebuttonstyle.ts
@@ -22,8 +22,8 @@ const classes = {
             'p-togglebutton-checked': instance.checked,
             'p-invalid': instance.invalid(),
             'p-disabled': instance.$disabled(),
-            'p-togglebutton-sm p-inputfield-sm': instance.size === 'small',
-            'p-togglebutton-lg p-inputfield-lg': instance.size === 'large',
+            'p-togglebutton-sm p-inputfield-sm': instance.size() === 'small',
+            'p-togglebutton-lg p-inputfield-lg': instance.size() === 'large',
             'p-togglebutton-fluid': instance.fluid()
         }
     ],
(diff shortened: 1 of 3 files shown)


gianluca-moro (approved):
I implemented these changes originally in PrimeNG, so I know they work ;)

dominicbachmann:
Hi @omurcanuslu thank you for the contribution! Can you please fix the merge conflicts? The PR should be ready to merge then :)

dominicbachmann (approved):

PR #1390, opened 2026-07-23
fix(#212): autocomplete optionLabel for string values correctly displayed

### LINKED ISSUE #212: [AutoComplete] optionLabel not applied to selected item if suggestions are strings
### Describe the bug

For AutoComplete components, we can use `optionLabel` to display a custom label in the suggestions and selected item.   
This works fine when the suggestions are objects but it does not work properly when the suggestions are just strings. In that case, the selected item does not display the label provided by `optionLabel` but just the value itself, while with objects, the selected item displays the label generated by `optionLabel`.

Image

<br/><br/>

Edit:   
After some initial investigation of the source code, I saw in https://github.com/primefaces/primeng/blob/993680d9202bce66570663d2c3c3eeda6860e56d/packages/primeng/src/autocomplete/autocomplete.ts#L860-L875 and https://github.com/primefaces/primeng/blob/993680d9202bce66570663d2c3c3eeda6860e56d/packages/primeng/src/autocomplete/autocomplete.ts#L921-L923 that the `optionLabel` is only applied to the selected item if the value is an object or if it's a string and the `optionValue` property is set.    
But this does not really make sense to me since why would you set `optionValue` if the value is a string??    
I think the logic in the `inputValue` signal computation is flawed. In my opinion, `optionLabel` should always be applied, no matter the type of the value. 

So, apparently there is a workaround to fix this issue: Add an `optionValue` function that just returns the item itself, aka:
```javascript
optionValue(item: any): any {
  return item;
}
```

### Pull Request Link

https://github.com/primefaces/primeng/pull/1074

### Reproducer

https://stackblitz.com/edit/dz8njim1-jg7bfq1g?file=src%2Fapp%2Fautocomplete-objects-demo.html

### Environment

PrimeNG 20.3.0, Angular 20, Firefox

### Angular version

20.0.4

### PrimeNG version

v20

### Browser(s)

Firefox

1. Go to https://stackblitz.com/edit/dz8njim1-jg7bfq1g?file=src%2Fapp%2Fautocomplete-objects-demo.html
2. The top AutoComplete uses an array of objects for the suggestions, while the bottom AutoComplete uses an array of strings
3. The optionLabel function is the same for both and just returns `test` 
4. When we select and item in both AutoCompletes, the top one still displays `test` (which is intended behaviour) while the bottom one displays the actual value instead of `test`

### Expected behavior

The selected item in AutoCopmlete component should always display the label generated by `optionLabel` no matter if the suggestions are objects or primitive types like strings.

> Migrated from `primefaces/primeng#19075` — originally by `@gianluca-moro` on 2025-11-10

gianluca-moro:
## Description

Correctly displays the optionLabel even if options are strings instead of just displaying the option itself.

## Related issues

Fixes #212 

## Type of change

- [ ] Documentation only

## Breaking changes

None

## Test plan

- [x] `npm run build`
- [x] `npm test`
- [x] `npm run lint`
- [x] Verified in the demo app (if applicable)

## Checklist

- [ ] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

## Additional context

Migrated from https://github.com/primefaces/primeng/pull/19076

### DIFF
--- a/packages/optimus-ui/src/autocomplete/autocomplete.ts
+++ b/packages/optimus-ui/src/autocomplete/autocomplete.ts
@@ -955,7 +955,7 @@ export class AutoComplete extends BaseInput<AutoCompletePassThrough> {
     }
 
     get optionValueSelected() {
-        return typeof this.modelValue() === 'string' && this.optionValue;
+        return typeof this.modelValue() === 'string' || this.optionValue;
     }
 
     chipItemClass(index) {


SoyDiego (approved):

dominicbachmann:
Hi @gianluca-moro thank you for your contribution. While reviewing this PR I found some more issues in the autocomplete component. I created https://github.com/openng-org/optimus-ui/pull/1608. Would you be open to review my PR?

Issue #57, opened 2026-06-29
Accessibility | Breadcrumb - separators aren't hidden in accessibility tree


geromegrignon:
### Describe the bug

While the list semantic is correct, the separators (chevron icon) should not be exposed as listitems. Also the chevron icons should not be exposed in the accessibility tree.

This is a duplicate if #13437, which is closed but partly not fixed.

<img width="2417" height="1440" alt="Image" src="https://github.com/user-attachments/assets/05983af9-3b49-4870-816d-7f939414987d" />

### Reason for not contributing a PR

- [x] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/github-ynsptzwi?file=src%2Fapp%2Fapp.component.html

### Environment

/

### Angular version

21.2.4

### PrimeNG version

v21

### Node version

24.13.0

### Browser(s)

Chrome

### Steps to reproduce the behavior

1.  Open the https://primeng.org/breadcrumb page
2. Inspect the first example for the Breadcrumb component
3. The lis used as separators show up in the the accessibility tree

### Expected behavior

The lis used as separators DO NOT show in accessibility tree.

> Migrated from `primefaces/primeng#19520` — originally by `@fgeierst` on 2026-04-07

geromegrignon:
*[Originally posted by `@github-actions[bot]` on 2026-04-07T13:55:23.000Z]*

Thanks for reporting this!

This issue appears similar to #13437 (Breadcrumb - separators aren't hidden from accessibility tree). Please check if that issue describes your problem.

- If **yes**, please add a thumbs up to that issue instead
- If **no**, please clarify how your issue differs

💼 Need priority assistance? Check out [Pro Support](https://primeng.org/support).
*This is an automated message.*

geromegrignon:
*[Originally posted by `@fgeierst` on 2026-04-07T13:56:34.000Z]*

While this is a duplicate, the original ticket is closed but the issue is not resolved.

dmorfav:
This one looks already fixed on `main`.

The `aria-hidden="true"` on both separator `<li>` came in with #1319 (`4367b5e`), merged on 2026-07-16 — after this was reported in April, so the report was accurate at the time.

Current accessibility tree for the Basic demo on the docs site (Chrome, `/breadcrumb`):

```
navigation
  list
    listitem → link ""
    listitem → link "Electronics"
[...]

PR #1406, opened 2026-07-26
fix(breadcrumb): give the icon-only home link an accessible name

### LINKED ISSUE #1403: [Bug]: Accessibility | Breadcrumb - home link has no accessible name when it renders an icon only
### Checklist

### What happened?

The `home` link exposes no accessible name when it renders as an icon only — which is the default, and the usual breadcrumb pattern. axe reports a `link-name` violation (serious, WCAG 2.4.4 / 4.1.2) and a screen reader announces just "link".

`homeAriaLabel` exists for exactly this. Its JSDoc reads *"Defines a string that labels the home icon for accessibility"*, but the input has no default value (`breadcrumb.ts:203`), so `[attr.aria-label]="homeAriaLabel"` resolves to `undefined` and the attribute is never rendered.

The component always renders an icon for `home`: `home.icon` when provided, otherwise a built-in `home` SVG (`breadcrumb.ts:44`). So unless the consumer passes `home.label` or `homeAriaLabel`, the link ends up nameless. The docs Basic demo reproduces it, declaring `home = { icon: 'pi pi-home' }`.

### Steps to reproduce

1. Open the Basic demo on the docs app (`/breadcrumb`)
2. Inspect the accessibility tree, or run axe / Lighthouse on the page

Accessibility tree in Chrome:

```
navigation
  list
    listitem → link ""          ← home, no accessible name
    listitem → link "Electronics"
    listitem → link "Computer"
    ...
```

axe:

```
link-name — Links do not have a discernible name
nav.p-breadcrumb > ol.p-breadcrumb-list > li.p-breadcrumb-home-item > a.p-breadcrumb-item-link
```

### Expected behavior

The home link has an accessible name out of the box, without the consumer having to know about `homeAriaLabel`.

Suggestion: default `homeAriaLabel` from the translation service instead of hardcoding an English string. There is no suitable key yet — the `Aria` interface has no `home` entry — so this would mean adding one (e.g. `aria.home: "Home"`). That overlaps with the default discussion in #51, so it may be worth settling both together.

The Basic demo probably deserves updating too, so the documented example is the accessible one.

### Actual behavior

The home link renders with no accessible name. It surfaces in the accessibility tree as `link ""`, and axe reports a `link-name` violation on it. A screen reader announces it as just "link", with no indication that it goes to the home page.

### Minimal reproduction

https://github.com/openng-org/optimus-ui/blob/main/apps/docs/doc/breadcrumb/basic-doc.ts#L29

### Library version

1.0.0-rc.1

### Angular version

21.2.18

### Node version

24.13.0

### Package manager

pnpm

### Browser(s)

Chrome

### Environment details

```markdown
optimus-ui `main` (`0d4ef80`), Angular 21.2.18, Chrome 150, axe-core 4.10.2
```

### Angular application setup

None

### Angular context

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional context

_No response_

rene-schakmann:
## Defect Fix

Fixes #1403

## Problem

Breadcrumb always renders an icon for the home item (either `home.icon` or the built-in SVG), but `homeAriaLabel` had no default. `[attr.aria-label]="homeAriaLabel"` therefore resolved to `undefined` and the attribute was never rendered.

With the default icon-only pattern (`home = { icon: 'pi pi-home' }`, as used by the Basic demo) the link ends up with no accessible name:

- axe reports a `link-name` violation (WCAG 2.4.4 / 4.1.2)
- screen readers announce only "link"
- the accessibility tree shows `link ""`

## Solution

The home link now falls back to a new `aria.home` translation key (`"Home"` by default) instead of hardcoded English, so the label is localizable through the locale configuration, consistent with how the other components resolve their aria labels.

Precedence:

1. `homeAriaLabel` when provided — unchanged behaviour
2. otherwise the `aria.home` translation, but only when `home.label` is not set

The second condition matters: when the home item has a visible label, that label already names the link, and adding an `aria-label` would override it (WCAG 2.5.3 Label in Name). So the fallback only kicks in for the icon-only case that actually lacks a name.

## Changes

- `packages/optimus-ui/src/api/translation.ts` — add `home?: string` to the `Aria` interface
- `packages/optimus-ui/src/config/optimus.ts` — default `aria.home` to `'Home'`
- `packages/optimus-ui/src/breadcrumb/breadcrumb.ts` — resolve the home link's `aria-label` through a `homeLinkAriaLabel` getter (both the `href` and the `routerLink` anchor)
- `packages/optimus-ui/src/breadcrumb/breadcrumb.spec.ts` — tests for the default translation, a configured translation, `homeAriaLabel` precedence, and the visible-label case
- `apps/docs/doc/breadcrumb/accessibility-doc.ts` — document how the home link is named
- `apps/docs/doc/configuration/locale/apidoc.ts` — document the `aria.home` key

## Verification

`ng test optimus-ui --include='**/breadcrumb/*.spec.ts'` → 89/89 passing (4 new).
Prettier and lint clean on the touched files.

The generated API docs (`apps/docs/doc/apidoc/index.json`, `apps/docs/public/llms/**`, `apps/docs/api-generator/typedoc.json`) are left untouched since they are produced by `build:docs:content`.

## Breaking Changes

None. Consumers that already set `homeAriaLabel` or `home.label` are unaffected; only the previously unnamed icon-only link gains a name.

---

### DIFF
--- a/apps/docs/doc/breadcrumb/accessibility-doc.ts
+++ b/apps/docs/doc/breadcrumb/accessibility-doc.ts
@@ -1,15 +1,20 @@
 import { Component } from '@angular/core';
+import { RouterModule } from '@angular/router';
 import { AppDocSectionText } from '@/components/doc/app.docsectiontext';
 
 @Component({
     selector: 'accessibility-doc',
     standalone: true,
-    imports: [AppDocSectionText],
+    imports: [RouterModule, AppDocSectionText],
     template: ` <app-docsectiontext>
         <h3>Screen Reader</h3>
         <p>
             Breadcrumb uses the <i>nav</i> element and since any attribute is passed to the root implicitly <i>aria-labelledby</i> or <i>aria-label</i> can be used to describe the component. Inside an ordered list is used where the list item
             separators have <i>aria-hidden</i> to be able to ignored by the screen readers. If the last link represents the current route, <i>aria-current</i> is added with "page" as the value.
+        </p>
+        <p>
+            The home item renders an icon, so when it has no <i>label</i> the link is named with the <i>homeAriaLabel</i> property. When <i>homeAriaLabel</i> is not defined either, the <i>aria.home</i> key of the
+            <a routerLink="/configuration" fragment="locale">locale</a> configuration is used as the default.
         </p>
 
         <h3>Keyboard Support</h3>
--- a/apps/docs/doc/configuration/locale/apidoc.ts
+++ b/apps/docs/doc/configuration/locale/apidoc.ts
@@ -323,6 +323,10 @@ import { AppDocSectionText } from '@/components/doc/app.docsectiontext';
                         <td>Navigation</td>
                     </tr>
                     <tr>
+                        <td>aria.home</td>
+                        <td>Home</td>
+                    </tr>
+                    <tr>
                         <td>aria.scrollTop</td>
                         <td>Scroll Top</td>
                     </tr>
--- a/packages/optimus-ui/src/api/translation.ts
+++ b/packages/optimus-ui/src/api/translation.ts
@@ -92,6 +92,7 @@ export interface Aria {
     previous?: string;
     next?: string;
     navigation?: string;
+    home?: string;
     scrollTop?: string;
     moveTop?: string;
     moveUp?: string;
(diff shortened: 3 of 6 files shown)


dmorfav:
@rene-schakmann, I checked this in an isolated working environment: I extracted the branch, compiled `optimus-ui`, and ran `ng test optimus-ui --include='**/breadcrumb/*.spec.ts'` in ChromeHeadless. The result was 89/89 tests passed, including the 4 new cases. The modified files were cleaner.

The precedence in `homeLinkAriaLabel` (`homeAriaLabel` → omit when `home.label` is visible → translate to `aria.home`) seems correct and respects the WCAG 2.5.3 naming convention. There's also no overlap with PR #1402: that one affects `menuitem` links, while this one affects `home` links, so the merge order doesn't matter.

It seems fine to me.

cc @geromegrignon, FYI since this follows up on the #57 thread.

geromegrignon:
thanks for your contribution @rene-schakmann  !

PR #1418, opened 2026-07-31
fix(tooltip): reset escaped content without innerHTML

### LINKED ISSUE #223: CSP Trusted Types fails for Tooltip
### Describe the bug

If using trusted types on Content Security Policy (CSP) on a PrimeNG app, I get the following error:

> Uncaught TypeError: Failed to set the 'innerHTML' property on 'Element': This document requires 'TrustedHTML' assignment.
    at _Tooltip.updateText (chunk-OJPK6ZHH.js?v=072743cd:547:34)
    at _Tooltip.create (chunk-OJPK6ZHH.js?v=072743cd:479:10)

### Pull Request Link

https://github.com/primefaces/primeng/pull/1080

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/github/primefaces/primeng-issue-template

### Environment

MacOS

### Angular version

20.3.7

### PrimeNG version

v20

### Node version

_No response_

### Browser(s)

Chrome

1. add a CSP configuration like this `default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self'; trusted-types angular angular#bundler; require-trusted-types-for 'script'` to your project (for example locally for testing in the `index.html` with 
```html
 <meta http-equiv="Content-Security-Policy"
      content="default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self'; trusted-types angular angular#bundler; require-trusted-types-for 'script'" />
```
2.  Open your app in the browser and look on the DevTools console

### Expected behavior

In that case an assignment on `innerHTML` is not necessary to reset the content.
The trusted types policy on the CSP should not error

> Migrated from `primefaces/primeng#19054` — originally by `@jase88` on 2025-11-01

jase88:
Fixes #223

## Problem

With a Trusted Types CSP (`require-trusted-types-for 'script'`), showing a tooltip throws:

> Uncaught TypeError: Failed to set the 'innerHTML' property on 'Element': This document requires 'TrustedHTML' assignment.

`Tooltip.updateText()` assigned an empty string to `innerHTML` purely to clear previously rendered text before appending the escaped text node. Trusted Types rejects any plain-string `innerHTML` assignment, including the empty one.

## Change

Clear the tooltip text with `textContent = ''` instead. Same effect (removes all child nodes), no Trusted Types violation.

The `escape: false` branch still assigns `innerHTML` — that path intentionally renders HTML and is out of scope here.

## Test

Added a unit test that makes `innerHTML` throw on the tooltip text element (mimicking a Trusted Types document) and asserts `updateText()` still replaces the previous content.

`ng test optimus-ui --include='**/tooltip/tooltip.spec.ts'` → 45/45 passing.

### DIFF
--- a/packages/optimus-ui/src/tooltip/tooltip.test.ts
+++ b/packages/optimus-ui/src/tooltip/tooltip.test.ts
@@ -486,6 +486,30 @@ describe('Tooltip', () => {
             expect(document.createTextNode).toHaveBeenCalledWith('Test content');
         });
 
+        it('should reset previous content without assigning innerHTML when escape is true', async () => {
+            const tooltipText = document.createElement('div');
+            tooltipText.appendChild(document.createTextNode('Previous content'));
+            Object.defineProperty(tooltipText, 'innerHTML', {
+                configurable: true,
+                get: () => '',
+                set: () => {
+                    // Mimics a Trusted Types enabled document rejecting plain string assignments
+                    throw new TypeError("Failed to set the 'innerHTML' property on 'Element': This document requires 'TrustedHTML' assignment.");
+                }
+            });
+
+            tooltipDirective.tooltipText = tooltipText;
+            tooltipDirective.setOption({ escape: true });
+            tooltipDirective.setOption({ tooltipLabel: 'New content' });
+            fixture.changeDetectorRef.markForCheck();
+            await fixture.whenStable();
+            fixture.detectChanges();
+
+            expect(() => tooltipDirective.updateText()).not.toThrow();
+            expect(tooltipText.childNodes.length).toBe(1);
+            expect(tooltipText.textContent).toBe('New content');
+        });
+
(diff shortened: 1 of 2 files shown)


jase88:
@geromegrignon @dominicbachmann 
Recreated this MR from PrimeNG PR.
Would love to see this being reviewed and will port more of my pending PRs if this lands

dominicbachmann (changes requested):
Hi @jase88 thank you for this fix! Can you please resolve the conflicts and also add the test to the vitest suite?

geromegrignon:
Just a follow-up on previous message: Karma has ben removed so the tests should be ported from Karma to Vitest (from the spec.ts file to the test.ts one)

jase88:
Thank you! Rebased and should be good to go

dominicbachmann (approved):

PR #1442, opened 2026-08-05
fix(tooltip): ignore mouseenter that follows no mouseleave

### LINKED ISSUE #950: Tooltip: Persists when dropdown is closed and ripple enabled
### Describe the bug

When ripple is enabled, and a Tooltip is added to a DropdownComponent, the tooltip reappears when an item from the dropdown is selected. 

### Environment

Chrome
NOTE: The issue does not appear in Firefox

### Reproducer

https://stackblitz.com/edit/primeng-dropdown-demo-xsrf3t?file=src/app/app.module.ts

### Angular version

15

### PrimeNG version

15

### Build / Runtime

Angular CLI App

### Language

TypeScript

18

### Browser(s)

Chrome 108

1. Set up a project where `PrimeNGConfig.ripple` is set to true as part of the app initializer.
2. Set up a `<p-dropdown pTooltip="tooltip"></p-dropdown>` (with valid options, etc)
3. Click on the dropdown to open it
4. Click on one of the dropdown items

The tooltip disappears, then reappears. Looking at the code, it seems that an `onMouseEnter` event is fired, causing the tooltip to reappear. The tooltip disappears the next time you hover on the dropdown.

### Expected behavior

Tooltip should not appear when dropdown item is clicked and mouse is not on dropdown any longer.

> Migrated from `primefaces/primeng#12412` — originally by `@dcbroad3` on 2022-12-21

rene-schakmann:
Fixes #950 — tooltip reappears after selecting an item, with ripple enabled, in Chrome.

## Root cause

The overlay of a `p-select` / `p-dropdown` is rendered **inside** the tooltipped host (`appendTo` defaults to `'self'`), so the option `<li>` the pointer sits on is a descendant of the element `Tooltip` listens on.

Selecting an option detaches that `<li>` — together with the `<span class="p-ink">` of its running ripple animation. Chrome then re-runs its hover computation for the detached subtree and dispatches a **fresh `mouseenter` at the host**: no pointer movement, and no `mouseleave` in between. `Tooltip.onMouseEnter` only guarded on `!this.container && !this.showTimeout`, both true right after the click hid the tooltip, so it re-showed the tooltip that the very same click had just dismissed — leaving it on screen while the pointer is nowhere near the host.

Firefox only recomputes hover on real pointer movement, which is why the report is Chrome-only. Removing the ripple ink from the DOM makes the spurious enter disappear as well, matching "only when ripple is enabled" from the report.

Instrumented event log from the reproduction (pointer stationary throughout):

```
t=7900  click       on host (option click bubbles up) -> deactivate(), tooltip hidden
t=7912  mouseenter  on host, x/y unchanged, no mouseleave  -> activate(), tooltip back
```

## The fix

An enter that is not preceded by a leave is not a new hover. `Tooltip` now tracks whether the pointer is already inside the host and ignores those re-entrant events. Genuine hover, leave/re-enter, focus and touch paths are untouched.

## Verification

Reproduced in the environment named in the issue — Angular 15 + PrimeNG 15, ripple on via `PrimeNGConfig`, `pTooltip` on `p-dropdown`, Chrome 108 in Docker — driven over WebDriver with real pointer events and **no pointer movement after the item click**.

| Before | After |
|---|---|
| before | after |

"Rome" is selected in both; on the left the tooltip is stranded next to the closed dropdown.

- 6 new specs in `tooltip.spec.ts` covering the invariant end to end. 3 of them fail without the fix and pass with it.
- `pnpm run test:unit`: 7210 passing. The single failure is `ScrollTop Performance should handle rapid scroll events efficiently`, which is timing-sensitive under full-suite load and passes when `scrolltop.spec.ts` runs on its own — unrelated to this change.
- `pnpm run format:check` clean. Note: `pnpm run lint` fails on `main` as well (`eslint --ignore-path` is not valid with the flat config), so it could not be used as a gate here; `eslint` run directly on the changed files reports no errors.
- Checked in the docs app on Chrome 150 that hover, leave/re-hover and select-an-option all behave.

Co-authored by Claude.

### DIFF
--- a/packages/optimus-ui/src/tooltip/tooltip.test.ts
+++ b/packages/optimus-ui/src/tooltip/tooltip.test.ts
@@ -281,6 +281,97 @@ describe('Tooltip', () => {
         });
     });
 
+    describe('Re-entrant Mouse Enter', () => {
+        let fixture: ComponentFixture<TestBasicTooltipComponent>;
+        let component: TestBasicTooltipComponent;
+        let tooltipDirective: Tooltip;
+        let inputElement: HTMLElement;
+
+        // Chrome dispatches a fresh mouseenter on the host when the element under the pointer is
+        // detached - for instance an overlay rendered inside the host closing on click. The pointer
+        // never moved, so no mouseleave precedes it. See issue #950.
+        const enter = () => inputElement.dispatchEvent(new MouseEvent('mouseenter'));
+        const leave = () => inputElement.dispatchEvent(new MouseEvent('mouseleave'));
+        const click = () => inputElement.dispatchEvent(new MouseEvent('click'));
+
+        beforeEach(() => {
+            fixture = TestBed.createComponent(TestBasicTooltipComponent);
+            component = fixture.componentInstance;
+            fixture.detectChanges();
+
+            const debugElement = fixture.debugElement.query(By.directive(Tooltip));
+            tooltipDirective = debugElement.injector.get(Tooltip);
+            inputElement = component.inputElement.nativeElement;
+        });
+
+        afterEach(() => {
+            tooltipDirective.deactivate();
+        });
+
+        it('should activate on the first mouse enter', () => {
+            const activateSpy = vi.spyOn(tooltipDirective, 'activate');
+
+            enter();
+
+            expect(activateSpy).toHaveBeenCalledTimes(1);
+            expect(document.querySelector('.p-tooltip')).toBeTruthy();
+        });
+
+        it('should ignore a mouse enter that is not preceded by a mouse leave', () => {
+            enter();
+            tooltipDirective.deactivate();
+
+            const activateSpy = vi.spyOn(tooltipDirective, 'activate');
+            enter();
+
+            expect(activateSpy).not.toHaveBeenCalled();
+        });
+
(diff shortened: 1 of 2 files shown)


geromegrignon (changes requested):
Please port your test to the `test.ts` file: we got rid of karma for the benefit of Vitest only (that's why you have a conflict hee too).

rene-schakmann:
Done — ported to `tooltip.test.ts` and rebased onto `main`, so the karma `tooltip.spec.ts` (deleted upstream) is gone from the diff and the conflict is resolved.

The new `Re-entrant Mouse Enter` block uses the vitest idioms of the surrounding file (`vi.spyOn`, `toBe(true)`), no jasmine leftovers.

```
✓ optimus-ui (chromium) src/tooltip/tooltip.test.ts (50 tests) 2064ms

 Test Files  1 passed (1)
      Tests  50 passed (50)
```

Sanity-checked that the tests actually cover the regression: with the `pointerInside` guard removed from `onMouseEnter`, 3 of the 6 new tests fail; with it in place all 50 pass.

Co-authored by Claude.

geromegrignon:
Thanks @rene-schakmann, merged!

geromegrignon (approved):

PR #1446, opened 2026-08-06
fix(docs): stop claiming ng add runs the PrimeNG migration

### LINKED ISSUE #1444: [Docs]: Does `ng add` perform the migration?
### Documentation is

- [ ] Missing
- [ ] Outdated
- [x] Confusing
- [ ] Typo or formatting error
- [ ] Not sure

### Documentation location

https://github.com/openng-org/optimus-ui/blob/820a69ad8915c3447704081df9c4b94dbbeedec2/README.md

### What is wrong or missing?

README claims:

> Run the same command in an existing PrimeNG v21 workspace. It detects the primeng dependency and runs the migration instead of a fresh install — packages swapped, imports rewritten, then a report of anything it could not handle automatically, with file and line numbers.

Migration guide says:

> Note that `ng add` does not run the migration. It only sets up Optimus UI in a fresh project and makes no changes when primeng is detected. The schematic can be re-run at any time, for example after pulling in unmigrated code, and accepts a couple of flags. --skip-install skips the package install task and --force bypasses the PrimeNG v21 version check.

### Suggested improvement

The code indicates that the migrations is **not** performed when running the `ng add` command. So my suggestion is to edit the README and add the `ng generate @openng/optimus-ui:migrate-from-primeng
` command under the migration section.

### Screenshots

_No response_

### Contribution

- [x] I plan to submit a pull request for this documentation fix

Ionaru:
Fixes #1444.

## The problem

The README and the migration guide contradict each other, and the README is the one that is wrong.

`ng add` stopped migrating PrimeNG workspaces in 851df785bf ("refactor(schematics): separate ng add and migration concerns", #1396, 2026-07-25). Since then `ngAdd` detects the dependency and bails:

```ts
if (hasPrimeng(tree)) {
    context.logger.warn('primeng detected — ng-add only sets up Optimus UI in new projects, so no changes were made.\n' +
        'To migrate this workspace, run: ng generate @openng/optimus-ui:migrate-from-primeng');
    return tree;
}
```

The README's "Migrating from PrimeNG" section was written in 2fde504159 three days *after* that removal and never revisited, so it has described non-existent behaviour since it landed. A reader following it runs `ng add`, sees a warning, and gets nothing migrated.

This is not a v1/v2 split. Auto-migration only ever shipped in `1.0.0-rc.0` and `1.0.0-rc.1`; `schematics/ng-add/index.js` is byte-identical across `1.0.0-rc.2`, `1.0.0`, `1.0.1` (`latest`) and `2.0.0-rc.0` (`rc`).

## Changes

| File | Was | Now |
| --- | --- | --- |
| `README.md` | "It detects the `primeng` dependency and runs the migration instead of a fresh install" | `npm install` + `ng generate @openng/optimus-ui:migrate-from-primeng` |
| `doc/faq/migration-doc.ts` | "Run `ng add` ... and it runs the full migration" | Same correction |
| `doc/philosophy/trademark-doc.ts` | "it is why `ng add` ships a migration schematic" | "it is why we ship a migration schematic" |
| `doc/installation/ngadd-doc.ts` | "it makes no changes at all" | "makes no changes to your code", plus "It does not run that schematic for you" |
| `doc/installation/prerequisites-doc.ts` | "the same `ng add` command detects that case for you" | Names `migrate-from-primeng` as a separate command |

`doc/migration/primeng/automated-doc.ts` was already correct and is unchanged.

On `ngadd-doc.ts`: "makes no changes at all" is true of the schematic but not of the command. `packages/optimus-ui/package.json` declares `"ng-add": { "save": "dependencies" }`, so the CLI installs and keeps `@openng/optimus-ui` before the schematic runs. A PrimeNG user who runs `ng add` ends up with both `primeng` and an unwired `@openng/optimus-ui` in dependencies. Hence the narrower wording.

## Verification

- `pnpm run format:check` passes.
- llms mirrors regenerated with `pnpm --filter docs build:llm-docs`; the only change beyond the corrected copy is the `generatedAt` date.
- `pnpm run lint` currently fails on `main` for unrelated reasons (the script passes `--ignore-path`, which eslint's flat config no longer accepts). Not touched here.

## Follow-ups

Two behavioural problems turned up while confirming what `ng add` actually does. Both are out of scope here, filed separately:

- #1447: `ng add` on a PrimeNG workspace prompts for a theme, installs the package, then makes no changes and exits 0.
- #1448: `hasPrimeng` misses `peerDependencies` and friends, and `addDefaultImport` then emits a duplicate `Aura` import that fails to compile.


SoyDiego (approved):

geromegrignon (approved):

geromegrignon:
Thanks for your contribution @Ionaru !

PR #1452, opened 2026-08-08
feat(types): export the severity types

### LINKED ISSUE #1420: [Feature]: Export a shared Severity type
### Checklist

- [ ] I have searched the existing issues and have not found a duplicate
- [x] I have read the README and documentation

## Description

Many Optimus UI components accept a `severity` property with a fixed set of string literal values, for example:

```ts
'info' | 'success' | 'warn' | 'danger' | 'secondary' | 'contrast' | 'help'
```

However, this type is not currently exported as a public API.

As a result, developers have to duplicate the union type in their own code when returning a severity from helper methods or services:

```ts
getSeverity(status: Status): 'info' | 'success' | 'warn' | 'danger' | 'secondary' {
    // ...
}
```

Without an explicit return type, TypeScript often infers `string`, which is not assignable to component inputs expecting the specific literal union.

## Proposed solution

Export a shared type from the public API, for example:

```ts
export type Severity =
    | 'info'
    | 'success'
    | 'warn'
    | 'danger'
    | 'secondary'
    | 'contrast'
    | 'help';
```

Components such as `Tag`, `Badge`, `Message`, `Toast`, and others could then reuse this type internally, while application developers could also import it:

```ts
import type { Severity } from '@optimus-ui/core';

getSeverity(status: Status): Severity {
    // ...
}
```

## Benefits

* Eliminates duplication of the string literal union.
* Provides a single source of truth for supported severity values.
* Improves maintainability if new severity values are added in future releases.
* Enhances the TypeScript developer experience with better type safety and autocompletion.

### Proposed solution

What do you think?

### Proposed API or usage

```typescript

```

### Alternatives considered

_No response_

### How important is this feature to you?

Nice to have

### Contribution

- [x] I am willing to open a pull request for this feature

### Additional context

_No response_

jase88:
Fixes #1420

The severity union was not public API, so apps had to duplicate it. Only `ButtonSeverity` was named — the other twelve declarations were inline unions that had drifted into three distinct value sets.

Each set is now a documented type in the types folder of the component that owns it:

| Type | Values |
| --- | --- |
| `BadgeSeverity` | `success` `info` `warn` `danger` `secondary` `contrast` — also used by Tag, OverlayBadge, FileUpload |
| `ButtonSeverity` | the six above plus `help` `primary` |
| `MessageSeverity` | as Badge, but `error` instead of `danger` |

All thirteen declarations reference these now. No component accepts a different set of values than before, apart from the `badgeSeverity` fix below, and the existing specs pass untouched.

The types hold only the literal values; `null` and `undefined` are composed at the use sites, so `getSeverity(): BadgeSeverity` cannot return null. This narrows `ButtonSeverity`, which had baked them in — compile-time only, and every input still accepts the empty values.

Two bugs fall out of the wiring:

- `badgeSeverity` was typed with the button severities on both `Button` and `ButtonProps`, but it is bound to the badge, which styles only the six badge values — `help` and `primary` were assignable and rendered an unstyled badge. Now `BadgeSeverity`. `ButtonProps` also said `'warning'` where everything else says `'warn'`. `strictTemplates` is off, so nothing flagged either.
- The private `Badge.setSeverity` parameter omitted `'secondary'` and `'contrast'`, which the public input accepts.

`ToastMessageOptions.severity` stays `string` — narrowing it would break callers, worth deciding separately along with the `'error'` vs `'danger'` split.

Verified: build clean, 7205 unit tests passing with no spec edits, apidoc renders all three with their literal lists, `format:check` clean.


packages/optimus-ui/src/types/button/button.types.ts
```diff
@@ -110,18 +110,22 @@ export interface ButtonProps {
...
     autofocus?: boolean | undefined;
     variant?: string | undefined;
 }
 
-export type ButtonSeverity = 'success' | 'info' | 'warn' | 'danger' | 'help' | 'primary' | 'secondary' | 'contrast' | null | undefined;
+/**
+ * Severity levels of the button.
+ * @group Types
+ */
+export type ButtonSeverity = 'success' | 'info' | 'warn' | 'danger' | 'help' | 'primary' | 'secondary' | 'contrast';
```
dominicbachmann: Removing null and undefined from the type is a breaking change and therefore has to be released in a new major version. I think for now we could remove this part so that we can already export the new types and create a new issue tagged for v3 for cleaning up the ButtonSeverity type.

dominicbachmann (changes requested):

packages/optimus-ui/src/types/button/button.types.ts
jase88: Didn't really think it through. It shouldn't be a breaking change anymore

dominicbachmann (approved):

dominicbachmann:
@jase88 Thank you for this nice contribution!

dominicbachmann:
Created https://github.com/openng-org/optimus-ui/issues/1607

PR #1456, opened 2026-08-08
fix(deps): refresh lockfile to patched nanoid

### LINKED ISSUE #1455: CI fails on `pnpm audit`: nanoid advisory GHSA-2v37-7h3g-55p8 reaches the prod graph
CI fails on every PR and push to `main`.

`pnpm run build` → `build:check` → `security:check` → `pnpm audit --prod --audit-level high`, which reports:

```
high  nanoid: custom generators can loop indefinitely when given non-integer sizes
Vulnerable versions  <3.3.17
Patched versions     >=3.3.17
https://github.com/advisories/GHSA-2v37-7h3g-55p8
```

`nanoid` comes in via `postcss`. It counts as prod because `apps/docs` lists `postcss`, `autoprefixer` and `tailwindcss` under `dependencies` instead of `devDependencies`. The docs app is never published, so nothing vulnerable ships.

Fix: no manifest change needed. `postcss` allows `nanoid: ^3.3.11`, so the lockfile just needs refreshing — `3.3.12` and `3.3.16` collapse to `3.3.17` and the audit passes.

Follow-up: move those three to `devDependencies` in `apps/docs` so build tooling stops showing up in the prod audit.

jase88:
Fixes #1455

`pnpm audit --prod --audit-level high` flagged nanoid `<3.3.17` ([GHSA-2v37-7h3g-55p8](https://github.com/advisories/GHSA-2v37-7h3g-55p8)). It runs via `pnpm run build`, so CI failed on every PR and push to `main`.

`postcss` already allows `nanoid: ^3.3.11`, so no manifest change is needed — just a lockfile refresh. `3.3.12` and `3.3.16` collapse to `3.3.17`.

Two things reviewers should know:

- the regen also refreshes resolutions for ~126 other transitive packages
- `packageManager` is bumped to the pnpm version that produced the lockfile, which is also what CI will now use

Second commit is fallout from the first: the regen moved prettier `3.8.3` → `3.9.6` (allowed by `^3.0.0`), which reformats empty `for` update clauses and comment-only objects. Six files, whitespace only.

Both checks pass locally: `format:check` clean, audit reports `No known vulnerabilities found`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,5 +81,5 @@
     "publishConfig": {
         "access": "public"
     },
-    "packageManager": "pnpm@11.13.1"
+    "packageManager": "pnpm@11.20.0+sha512.9a6f330a95b66446ea088faf1521405a8a01f07fde7124cc9958dfed52d4bb436737e65b08f85f37b46fcba375092558ac51262b816844b22f63406ed166bfee"
 }
--- a/packages/optimus-ui-utils/src/object/methods/deepEquals.ts
+++ b/packages/optimus-ui-utils/src/object/methods/deepEquals.ts
@@ -18,7 +18,7 @@ function _deepEquals(obj1: unknown, obj2: unknown, visited: WeakSet<object> = ne
     if (arrObj1 && arrObj2) {
         length = obj1.length;
         if (length != obj2.length) return false;
-        for (i = length; i-- !== 0; ) if (!_deepEquals(obj1[i], obj2[i], visited)) return false;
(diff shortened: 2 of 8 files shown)


jase88:
@geromegrignon would be great to get a review

geromegrignon:
Thanks for the proposal.
We just added dependabot and it'll be better to update deps individually rather than having multiple concerns at once.

PR #1458, opened 2026-08-08
fix(schematics): enhance project detection

### LINKED ISSUE #1448: [Bug]: hasPrimeng misses several dependency fields, and ng add then emits a duplicate theme import
### Checklist

### What happened?

Found while investigating #1444. Two defects that compound into a workspace that no longer compiles.

**1. Detection is incomplete.** `hasPrimeng` (`schematics/utils/package-json.ts:62-73`) only inspects `dependencies` and `devDependencies`:

```ts
if ('primeng' in (pkg.dependencies ?? {}) || 'primeng' in (pkg.devDependencies ?? {})) {
    found = true;
}
```

A real PrimeNG project is missed when `primeng` is declared only in `peerDependencies` (normal for a library package inside a monorepo), only in `optionalDependencies`, only via `resolutions` / `pnpm.overrides`, installed but undeclared, or declared in a `package.json` under `dist/` (excluded by `SKIP_DIRS`). In those cases the guard passes and `freshSetup` runs on a PrimeNG workspace.

**2. The default-import guard checks the wrong thing.** `addDefaultImport` (`schematics/utils/app-config.ts:11`) treats an import as already present only when the module specifier *and* the binding name both match:

```ts
statement.moduleSpecifier.text === module && statement.importClause?.name?.text === name
```

It never checks whether the binding name alone is already taken. PrimeNG's theme presets use the same names as ours (`Aura`, `Lara`, `Material`, `Nora`), so on a workspace that already has `import Aura from '@primeuix/themes/aura'` it happily adds a second `import Aura`, and the file stops compiling.

### Steps to reproduce

1. Take an Angular v21 app that uses PrimeNG with the Aura preset, so `app.config.ts` contains `import Aura from '@primeuix/themes/aura';` and `providePrimeNG({ theme: { preset: Aura } })`.
2. Declare `primeng` in `peerDependencies` rather than `dependencies`.
3. Run `ng add @openng/optimus-ui`.
4. Open `src/app/app.config.ts` and run `tsc`.

Verified against `main` (820a69ad89) using the repo's own test harness (`SchematicTestRunner` over a real `@schematics/angular` v21 `workspace` + `application` tree), then type-checking the emitted file.

### Expected behavior

Either `primeng` is detected and `ng add` bails as designed, or, if the schematic does proceed, it does not emit an import that collides with an existing binding.

### Actual behavior

`ng add` reports success and writes a file that does not compile:

```ts
import Aura from '@primeuix/themes/aura';
import { providePrimeNG } from 'primeng/config';
import { ApplicationConfig, provideBrowserGlobalErrorListeners } from '@angular/core';
import { provideRouter } from '@angular/router';

import { routes } from './app.routes';
import { provideOptimus } from '@openng/optimus-ui/config';
import Aura from '@openng/optimus-ui-themes/aura';

export const appConfig: ApplicationConfig = {
  providers: [providePrimeNG({ theme: { preset: Aura } }),
    provideBrowserGlobalErrorListeners(),
    provideRouter(routes), provideOptimus({ theme: { preset: Aura } })
  ]
};
```

`tsc` on that output:

```
app.config.ts:1 error TS2300: Duplicate identifier 'Aura'.
app.config.ts:8 error TS2300: Duplicate identifier 'Aura'.
```

Note also that both `provideOptimus` and `providePrimeNG` now reference the same `Aura` binding, so even after resolving the collision the preset wiring is wrong.

### Minimal reproduction

No external repro needed, reproducible directly against this repository. See the steps above.

### Library version

1.0.1, 2.0.0-rc.0, and `main` at 820a69ad89

### Angular version

21.0.0

### Node version

24.17.0

### Package manager

pnpm

### Browser(s)

Other

### Environment details

- OS: macOS
- Node: 24.17.0

### Angular application setup

Standalone components

### Relevant log output

```shell
info: Added provideOptimus with the Aura preset to /src/app/app.config.ts.
```

### Additional context

Suggested fixes, independent of each other:

- Widen `hasPrimeng` to cover `peerDependencies` and `optionalDependencies`, and consider `resolutions` / `pnpm.overrides`. `hasPrimeflex` (`package-json.ts:80-91`) has the same narrow check and would want the same treatment.
- Make `addDefaultImport` collision-aware: if the binding name is already bound to a different module, either alias the new import (`import AuraPreset from '@openng/optimus-ui-themes/aura'`) or fall back to the manual-instructions path rather than emitting broken code.

The second fix is the more important of the two, since it protects every path into `addDefaultImport`, not just the ones where detection failed.

geromegrignon:
## Description

Fix migration schematic detections.

## Related issues

Fixes #1448

## Type of change

- [ ] Documentation only

## Breaking changes

None

## Test plan

- [x] `npm run build`
- [x] `npm test`
- [x] `npm run lint`
- [ ] Verified in the demo app (if applicable)

## Checklist

### DIFF
--- a/packages/optimus-ui/schematics/ng-add/index.ts
+++ b/packages/optimus-ui/schematics/ng-add/index.ts
@@ -132,11 +132,25 @@ async function wireProvideOptimus(tree: Tree, context: SchematicContext, options
         // `external()` only emits named imports, and `@openng/optimus-ui-themes/aura` only has a
         // default export — add it ourselves.
         const original = tree.read(wiredFile)!.toString();
-        const updated = addDefaultImport(original, preset, module);
+        const imported = addDefaultImport(original, preset, module);
+        let updated = imported.text;
+        if (imported.bindingName !== preset) {
+            // The preset name is already taken in this file (e.g. `Aura` imported from another
+            // theme package), so the import went in under a different binding — retarget the
+            // `preset:` reference in the provider call addRootProvider just inserted. `external()`
+            // may itself have aliased the provideOptimus binding (provideOptimus_1, …), hence `\w*`.
+            const providerCall = new RegExp(`(provideOptimus\\w*\\(\\{\\s*theme:\\s*\\{\\s*preset:\\s*)${preset}(\\s*\\}\\s*\\}\\s*\\))`);
+            if (!providerCall.test(updated)) {
+                context.logger.warn(instructions);
+                return;
+            }
+            updated = updated.replace(providerCall, `$1${imported.bindingName}$2`);
+        }
         if (updated !== original) {
             tree.overwrite(wiredFile, updated);
         }
-        context.logger.info(`Added provideOptimus with the ${preset} preset to ${wiredFile}.`);
+        const bindingNote = imported.bindingName === preset ? '' : ` (bound as ${imported.bindingName})`;
+        context.logger.info(`Added provideOptimus with the ${preset} preset to ${wiredFile}${bindingNote}.`);
     } catch {
         context.logger.warn(instructions);
     }
--- a/packages/optimus-ui/schematics/ng-add/index.ts
+++ b/packages/optimus-ui/schematics/ng-add/index.ts
@@ -15,13 +15,16 @@ function themeImport(theme: Theme): { preset: string; module: string } {
     return { preset: theme, module: `${THEMES_PACKAGE}/${theme.toLowerCase()}` };
 }
 
-function manualInstructions(theme: Theme): string {
+function manualSteps(theme: Theme): string {
     const { preset, module } = themeImport(theme);
-    return `Could not find a providers array to update automatically. Finish the setup manually:
-  1. import { provideOptimus } from '@openng/optimus-ui/config';
+    return `  1. import { provideOptimus } from '@openng/optimus-ui/config';
   2. import ${preset} from '${module}';
   3. add provideOptimus({ theme: { preset: ${preset} } }) to your root providers
      (bootstrapApplication providers, or an NgModule's providers array).`;
+}
+
+function manualInstructions(theme: Theme): string {
+    return `Could not find a providers array to update automatically. Finish the setup manually:\n${manualSteps(theme)}`;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


packages/optimus-ui/schematics/utils/package-json.ts
Ionaru: Dependency detection was widened with `DIRECT_DEPENDENCY_SECTIONS`, but the actual action part here still uses only `['dependencies', 'devDependencies']`.

packages/optimus-ui/schematics/ng-add/index.ts
```diff
@@ -132,11 +132,25 @@ async function wireProvideOptimus(tree: Tree, context: SchematicContext, options
...
+        let updated = imported.text;
+        if (imported.bindingName !== preset) {
+            // The preset name is already taken in this file (e.g. `Aura` imported from another
+            // theme package), so the import went in under a different binding — retarget the
+            // `preset:` reference in the provider call addRootProvider just inserted. `external()`
+            // may itself have aliased the provideOptimus binding (provideOptimus_1, …), hence `\w*`.
+            const providerCall = new RegExp(`(provideOptimus\\w*\\(\\{\\s*theme:\\s*\\{\\s*preset:\\s*)${preset}(\\s*\\}\\s*\\}\\s*\\))`);
+            if (!providerCall.test(updated)) {
+                context.logger.warn(instructions);
+                return;
```
Ionaru, line 171: This check's bail-out leaves a potentially broken `provideOptimus({ theme: { preset: Aura } })` in the file. With `Aura` being whatever the conflicted package was.

Ionaru (changes requested):

packages/optimus-ui/schematics/utils/app-config.ts
```diff
@@ -1,19 +1,57 @@
...
+export function addDefaultImport(text: string, name: string, module: string): DefaultImportResult {
     const sourceFile = ts.createSourceFile('file.ts', text, ts.ScriptTarget.Latest, true);
+    const defaultImports = collectDefaultImports(sourceFile);
+
+    const existing = defaultImports.find((candidate) => candidate.module === module);
+    if (existing) {
+        return { text, bindingName: existing.bindingName };
+    }
+
+    const legacy = defaultImports.find((candidate) => mapModuleSpecifier(candidate.module) === module);
```
Ionaru, line 39: This retarget makes the failure mode worse in exactly the cases #1448 says are still unfixed.

The issue lists "installed but undeclared" and "declared in a `package.json` under `dist/`" as detection gaps, and this PR does not close either. In those workspaces `hasPrimeng` returns false, `freshSetup` runs, and a live PrimeNG app reaches here with `import Aura from '@primeuix/themes/aura'` and `providePrimeNG({ theme: { preset: Aura } })` still in the file.

Before this PR that produced a second `import Aura` and `TS2300: Duplicate identifier`, which is loud and immediately obvious. Now the specifier is rewritten in place and the binding name is kept, so `providePrimeNG` silently receives an Optimus preset object at runtime. It compiles, so nothing tells the user anything is wrong. That is the "even after resolving the collision the preset wiring is wrong" note from the issue, now without the compile error that used to surface it.

The new test for this path uses a fixture with no `providePrimeNG` call, so it does not cover this.

Suggestion: skip the in-place retarget when the file still contains a `providePrimeNG` call and fall back to the manual instructions.

PR #1483, opened 2026-08-13
fix(virtualscroller): fix row height CSS

### LINKED ISSUE #1457: [Bug]: TreeTable with Virtual Scroll changes row height dynamically
### Checklist

### What happened?

Using virtual scroll with a TreeTable changes the row height if you filter the table and only a few rows are visible.

In the following example, the rows have a height of 46px applied (via the `h-[46px]` class) and at the start, it looks as expected:

Image

But when I then filter so it only shows 2 rows, the rows now each take up half of the available height, which is obviously more than 46px:

Image

Disabling virtual scrolling fixes the issue, so it is caused by the virtual scroller. 

Changing scrollHeight to a fixed pixel height, rather than flex, did not fix the issue. 

Here is the table definition:
```html
<p-treetable
    [value]="files"
    [columns]="cols"
    [filterMode]="filterMode"
    [scrollable]="true"
    scrollHeight="flex"
    [virtualScroll]="true"
    [virtualScrollItemSize]="46"
>
...
  <ng-template #body let-rowNode let-rowData="rowData">
    <tr [ttRow]="rowNode" class="h-[46px]">
      ...
    </tr>
  </ng-template>
</p-treetable>
```

Furthermore, I also noticed that there is no documentation for the virtual scroll in the TreeTable documentation. 

### Steps to reproduce

1. Go to https://stackblitz.com/edit/pkn57xr2?file=src%2Fapp%2Ftreetable-filter-demo.ts
2. See that each row has a height of 46px
3. Filter so that only a few rows are visible (e.g., size=30)
4. You can now see that there are only two rows but each is way taller than before

### Expected behavior

The height of the table rows should remain no matter how many are displayed

### Actual behavior

The height of the rows changes depending on how many are visible

### Minimal reproduction

https://stackblitz.com/edit/pkn57xr2?file=src%2Fapp%2Ftreetable-filter-demo.ts

### Library version

2.0.0-rc.0

### Angular version

21.0.0

### Node version

22.22.3

### Package manager

npm

### Browser(s)

Firefox

### Environment details

```markdown
OS: Tested on Linux and MacOS
Browsers: Tested on Firefox and Chrome
OptimusUI: Tested on 2.0.0-rc.0 and 1.0.1

always same behaviour
```

### Angular application setup

Standalone components

### Angular context

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional context

_No response_

jaroban:
## Description

Min-height was making rows too high, but is needed elsewhere - in the horizontal scroller where it makes the items full-height. So the fix was to move the min-height from all scrollers to just the horizontal one.

## Related issues

Fixes #1457

## Type of change

- [ ] Documentation only

## Breaking changes

None

None

## Test plan

- [x] `npm run build`
- [ ] `npm test`
- [ ] `npm run lint`
- [x] Verified in the demo app (if applicable)

## Checklist

### DIFF
--- a/packages/optimus-ui/src/scroller/style/scrollerstyle.ts
+++ b/packages/optimus-ui/src/scroller/style/scrollerstyle.ts
@@ -15,7 +15,6 @@ const css = /*css*/ `
     position: absolute;
     top: 0;
     left: 0;
-    min-height: 100%;
     min-width: 100%;
     will-change: transform;
 }
@@ -54,6 +53,7 @@ const css = /*css*/ `
 
 .p-virtualscroller-horizontal > .p-virtualscroller-content {
     display: flex;
+    min-height: 100%;
 }
 
 .p-virtualscroller-inline .p-virtualscroller-content {


SoyDiego (approved):

geromegrignon:
Thanks for the contribution @jaroban

PR #1537, opened 2026-08-17
fix(cascadeselect): bind aria-label to the listLabel getter

### LINKED ISSUE #1453: [Bug]: CascadeSelect option list has no aria-label; template binds 'listlabel' instead of 'listLabel'
### Checklist

### What happened?

The CascadeSelect option list has no accessible name. The template binds `listlabel`, but the getter is `listLabel` (`cascadeselect.ts:372` vs `:766`), so the binding resolves to `undefined` and the attribute is dropped.

### Steps to reproduce

1. Open the overlay on https://www.openng.org/cascadeselect
2. Inspect `document.querySelector('p-cascadeselect ul[role="tree"]')`

### Expected behavior

```html
<ul role="tree" aria-label="Option List">
```

### Actual behavior

```html
<ul role="tree">
```

### Minimal reproduction

https://www.openng.org/cascadeselect

### Library version

1.0.0-rc.2

### Angular version

21.0.0

### Node version

24.15.0

### Package manager

pnpm

### Browser(s)

Chrome

### Environment details

- OS: macOS
- Node: 24.15.0

### Angular application setup

Standalone components

### Additional context

_No response_

AshwinSathian:
## Summary

The CascadeSelect option list's `[attr.aria-label]` binding referenced `listlabel`, a property that does not exist on the component. The binding resolved to `undefined`, so the `aria-label` attribute was dropped from the rendered `<ul role="tree">`. The component does define a `listLabel` getter (backed by the ARIA translation key), but the template never referenced it. This PR points the binding at `listLabel`, matching the pattern already used in Select, MultiSelect, and AutoComplete.

## Steps to reproduce (before this fix)

1. Open the overlay on the CascadeSelect docs page.
2. Inspect `document.querySelector('p-cascadeselect ul[role="tree"]')`.
3. The element has no `aria-label` attribute, so screen readers announce the option tree with no accessible name.

## Change

- `packages/optimus-ui/src/cascadeselect/cascadeselect.ts`: `listlabel` -> `listLabel`.
- `packages/optimus-ui/src/cascadeselect/cascadeselect.test.ts`: added a test asserting the rendered option list gets `aria-label="Option List"`. Verified it fails without the fix and passes with it.

Fixes #1453

### DIFF
--- a/packages/optimus-ui/src/cascadeselect/cascadeselect.test.ts
+++ b/packages/optimus-ui/src/cascadeselect/cascadeselect.test.ts
@@ -854,6 +854,22 @@ describe('CascadeSelect', () => {
             expect(ariaRequired === null || ariaRequired === 'false').toBe(true);
             expect(hiddenInput.nativeElement.getAttribute('aria-label')).toBeTruthy();
         });
+
+        it('should set an aria-label on the option list', async () => {
+            testComponent.options = mockCountries;
+            testFixture.changeDetectorRef.markForCheck();
+            await testFixture.whenStable();
+            testFixture.detectChanges();
+
+            const trigger = testFixture.debugElement.query(By.css('.p-cascadeselect-dropdown'));
+            trigger.nativeElement.dispatchEvent(new MouseEvent('mousedown', { bubbles: true }));
+            testFixture.changeDetectorRef.markForCheck();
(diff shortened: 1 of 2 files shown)


geromegrignon:
Duplicate PR.

PR #1538, opened 2026-08-17
fix(confirmdialog): add draggable to the Confirmation interface

### LINKED ISSUE #60: Draggable property missing in the Confirmation interface
ConfirmDialog API says there should be a draggable property in the properties param (https://primeng.org/confirmdialog), which if defined works, but is missing in the Confirmation interface which results in the following TS issue:
Object literal may only specify known properties, and draggable does not exist in type Confirmation

> Migrated from `primefaces/primeng#19516` — originally by `@rokamon1997` on 2026-04-03

AshwinSathian:
## Summary

`ConfirmationService.confirm()` already supports a `draggable` option at runtime: `ConfirmDialog`'s constructor copies every own key of the `Confirmation` object onto the component instance (`keys.forEach((key) => { this[key] = confirmation[key]; })`), and `draggable` is one of the component's own `@Input()`s, bound straight through to the underlying `p-dialog`. The `Confirmation` interface, however, never declared the field, so any typed caller doing `confirm({ draggable: false })` hits:

```
Object literal may only specify known properties, and 'draggable' does not exist in type 'Confirmation'.
```

This PR adds `draggable?: boolean` to the interface, next to the other dialog-behavior flags (`closable`, `modal`). No runtime code changes — the propagation path already existed.

## Change

- `packages/optimus-ui/src/api/confirmation.ts`: add `draggable?: boolean` to `Confirmation`.
- `packages/optimus-ui/src/confirmdialog/confirmdialog.test.ts`: added a test that calls `confirmationService.confirm({ draggable: false, ... })` and asserts it propagates to both the `ConfirmDialog` instance and the underlying `Dialog` instance. Verified it fails to type-check (`TS2353`) without the interface change and passes with it.

Fixes #60

### DIFF
--- a/packages/optimus-ui/src/api/confirmation.ts
+++ b/packages/optimus-ui/src/api/confirmation.ts
@@ -114,4 +114,9 @@ export interface Confirmation {
      * @defaultValue true
      */
     modal?: boolean;
+    /**
+     * Enables dragging to change the position of the dialog using its header.
+     * @defaultValue true
+     */
+    draggable?: boolean;
 }
--- a/packages/optimus-ui/src/confirmdialog/confirmdialog.test.ts
+++ b/packages/optimus-ui/src/confirmdialog/confirmdialog.test.ts
@@ -233,6 +233,24 @@ class TestConfirmationServiceComponent {
     }
 }
 
+// ConfirmDialog with a Confirmation-level draggable override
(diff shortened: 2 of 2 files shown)


geromegrignon:
Duplicate PR.

PR #1564, opened 2026-08-20
fix(confirmpopup): resolve options from the active confirmation

### LINKED ISSUE #1547: ConfirmPopup: confirmation keys copied onto the component are never cleared, so option() can serve stale button props
### Describe the bug

Follow-up from #1438 (review) — related to #40.

When a confirmation is shown, `ConfirmPopup` copies every key of the `Confirmation` object onto the component instance:

```ts
// confirmpopup.ts (requireConfirmation$ subscription)
this.confirmation = confirmation;
const keys = Object.keys(confirmation);

keys.forEach((key) => {
    this[key] = confirmation[key];
});
```

These copied keys are never cleared: `hide()` only resets `this.confirmation = null`, and the next confirmation only overwrites the keys it actually contains.

Since `option()` reads from the component instance (`this`) rather than from `this.confirmation`, the `[buttonProps]="getAcceptButtonProps()"` / `getRejectButtonProps()` bindings can serve **stale props from a previous `confirm()` call**:

1. `confirm({ acceptButtonProps: { severity: 'danger' }, ... })` → popup shows a danger accept button. ✅
2. Popup is closed.
3. `confirm({ message: 'Something else' })` (no `acceptButtonProps`) → the accept button **still** gets `{ severity: 'danger' }`, because `this.acceptButtonProps` was never cleared and `option('acceptButtonProps')` still returns it. ❌

This affects any `Confirmation` key resolved through `option()`/copied state when two confirmations with different shapes go through the same `<p-confirmpopup>` instance (the common case: one popup + one `ConfirmationService`).

Note: after #1438 the *labels* are safe, because the `acceptButtonLabel`/`rejectButtonLabel` getters read `this.confirmation?...` directly — this issue is about the remaining copied-key/`option()` paths (e.g. the `[buttonProps]` bindings).

### Suggested fix

Either:
- resolve `option()` against `this.confirmation` (like the label getters now do), or
- clear/reset the previously copied confirmation keys when hiding or when a new confirmation arrives.

`ConfirmDialog` uses the same copy-keys-onto-`this` pattern and is likely affected too — worth checking while fixing.

### Environment

- `packages/optimus-ui/src/confirmpopup/confirmpopup.ts`, observed on `main` (v2-rc.0)

rene-schakmann:
## Description

Fixes #1547.

`ConfirmPopup` and `ConfirmDialog` copied **every** key of the emitted `Confirmation` onto the component instance and never cleared them:

```ts
this.confirmation = confirmation;
Object.keys(confirmation).forEach((key) => { this[key] = confirmation[key]; });
```

`hide()` only flips visibility and `onContainerDestroy()`/`onDialogHide()` only null `this.confirmation`, so the copies survive. Since `option()` resolved against `this` rather than `this.confirmation`, the next `confirm()` inherited every key it did not itself set — the `[buttonProps]` bindings being the visible case.

It also **destroyed the components' own `@Input` defaults**. `ConfirmDialog` has real `@Input`s for `closable`, `modal`, `blockScroll`, `closeOnEscape`, `acceptVisible`, `rejectVisible`, `defaultFocus` and `position`; a single `confirm({ closable: false })` overwrote `this.closable` permanently, so every later confirmation on that instance was non-closable.

### The fix

- `option()` now resolves against the active `this.confirmation` first and falls back to the component instance, so the `@Input` defaults are what a confirmation falls back to — never a previous confirmation.
- The copy-keys-onto-`this` loop is removed from both components.
- `ConfirmDialog`'s `[position]` and `[dismissableMask]` bindings read the instance directly, so those `Confirmation` keys only worked *because* of the copy loop — they now go through `option()`.
- `ConfirmPopup.handleFocus()` read `this.defaultFocus` for the same reason and now uses `option('defaultFocus')`.

### Drive-by, same method

`option(name, k)` did a **top-level** lookup for the sub-key (`source[k]`) instead of looking `k` up on the named option, which left `ConfirmDialog`'s two `[ariaLabel]="option('acceptButtonProps', 'ariaLabel')"` bindings permanently `undefined`. It now reads `source[name]?.[k]`. Happy to split this out if you would rather keep the PR to the reported bug.

## Steps to reproduce

1. `confirm({ acceptButtonProps: { severity: 'danger' }, ... })` — accept button is danger-styled ✅
2. Close the popup.
3. `confirm({ message: 'Something else' })` — no `acceptButtonProps` at all.

**Before** — the accept button is *still* danger-red, inherited from step 1:

before - stale danger accept button

**After** — the accept button falls back to the default styling:

after - default styling restored

(The labels are correct in both shots — `acceptButtonLabel`/`rejectButtonLabel` already read `this.confirmation` directly, exactly as the issue notes. Only the `option()`-resolved paths were stale.)

Reproduced against the docs app with a third trigger that passes no button props; that scratch demo is not part of this PR.

## Type of change

- [ ] Documentation only

## Test plan

11 new tests across both components, added to the jasmine and vitest suites alike (`confirmpopup.spec.ts`/`.test.ts`, `confirmdialog.spec.ts`/`.test.ts`) under `describe('Confirmation State Isolation')`:

- `acceptButtonProps`/`rejectButtonProps` are not inherited by the next confirmation
- confirmation keys are not copied onto the component instance
- `closable` and `position` fall back to the `@Input` default after a confirmation that overrode them
- `defaultFocus` resolves from the confirmation, then falls back to the input
- `option(name, k)` reads the sub-key off the named option

All 11 fail on `main` and pass with the fix.

- [x] `pnpm run test:unit` (jasmine/karma, full suite)
- [x] `pnpm run test:vitest` (full suite)
- [x] `pnpm run format:check`
- [x] Verified in the demo app (screenshots above, plus the ConfirmDialog position demo as a regression check on the changed template bindings)

## Checklist

- [x] Public API changes documented; breaking changes called out

---

Co-authored by Claude (Claude Code). The trailer is in the PR description rather than the commit message because commitlint's `no-ai-co-author` rule rejects AI `Co-authored-by` trailers.


packages/optimus-ui/src/confirmdialog/confirmdialog.ts
geromegrignon: Change `rejectIcon` to `option('rejectIcon')`

geromegrignon: Change `acceptIcon` to `option('acceptIcon')`

geromegrignon (changes requested):
Besides the two comments, please rollback spec.ts changes.
They were pretty welcome and appreciated but i just merged the karma test suite removal.

geromegrignon (approved):

geromegrignon:
Thanks @rene-schakmann, merged!

PR #1565, opened 2026-08-20
fix(password): commit the cleared value under updateOn blur

### LINKED ISSUE #945: Component: The component<p-password>uses FormControl to set updateOn: 'blur' will not trigger form validation
### Describe the bug

N_IS}S I6EX8` $ O76)902

https://user-images.githubusercontent.com/35752648/215780559-54a0e080-febc-40ac-9a0d-0582cf28e223.mp4

See the form element "new password". I use a responsive form to configure the updateOn: blur event. I hope to prompt for verification information when blur. However, the<p-password>is triggered by the change event. The blur event does not take effect。

I have consulted some netizens who think the onInputBlur and onInput methods in the source code call onModelTouched at the wrong time.

In onInputBlur, netizens think it should be called as follows
_5P1$_JB IW)}5)C}$XSR

### Environment

..

### Reproducer

_No response_

### Angular version

15.0.4

### PrimeNG version

15.0.1

### Build / Runtime

Angular CLI App

### Language

TypeScript

v18.13.0

### Browser(s)

Chrome 109.0.5414.120

_No response_

### Expected behavior

_No response_

> Migrated from `primefaces/primeng#12584` — originally by `@AlanLuo-dev` on 2023-01-31

rene-schakmann:
## Description

Fixes #945.

With `updateOn: 'blur'`, Angular does **not** commit a value when the accessor calls `onModelChange` — it only stages it. The commit happens inside the callback registered through `registerOnTouched` ([`setUpBlurPipeline`](https://github.com/angular/angular/blob/main/packages/forms/src/directives/shared.ts)):

```ts
function setUpViewChangePipeline(control, dir) {
  dir.valueAccessor.registerOnChange(newValue => {
    control._pendingValue = newValue;
    control._pendingChange = true;
    control._pendingDirty = true;
    if (control.updateOn === 'change') updateControl(control, dir);
  });
}

function setUpBlurPipeline(control, dir) {
  dir.valueAccessor.registerOnTouched(() => {
    control._pendingTouched = true;
    if (control.updateOn === 'blur' && control._pendingChange) updateControl(control, dir);
    if (control.updateOn !== 'submit') control.markAsTouched();
  });
}
```

`Password.clear()` staged the cleared value and wrote it to the view, but never marked the control as touched:

```ts
clear() {
    this.value = null;
    this.onModelChange(this.value);   // stages null only
    this.writeValue(this.value);      // clears the visible input
    this.onClear.emit();
}
```

The clear icon sits outside the input, so clicking it produces no blur and nothing ever commits. **The field renders empty while the control still holds the previous password and still reports `VALID`** — a form submitted at that point sends the stale value, and `required` never fires.

### The fix

`clear()` now calls `onModelTouched()` after the model change, which is also what `Select.clear()` does. Under `updateOn: 'change'` the behaviour is unchanged (the value is already committed; the control is additionally marked touched, consistent with the other components).

### Note on the originally reported symptom

The keystroke half of the original report ("validation is triggered by the change event, blur does not take effect") no longer reproduces on this codebase: `onModelTouched()` is called only from `onInputBlur()` — upstream PrimeNG 15 also called it from `onInput()`, which committed the pending value on every keystroke. That part is already fixed here; I verified it in the browser and pinned it with tests so it cannot regress.

## Steps to reproduce

```ts
form = new FormGroup({
    value: new FormControl('', {
        validators: [Validators.required, Validators.minLength(8)],
        updateOn: 'blur'
    })
});
```

```html
<p-password formControlName="value" [feedback]="false" [showClear]="true" [toggleMask]="true" />
```

1. Type `secret12`
2. Blur the field — `control.value` becomes `"secret12"`, status `VALID` ✅
3. Click the clear icon

**Before** — the input is empty but the control still holds `"secret12"` and is `VALID`:

before - cleared input, stale control value

**After** — clearing commits, the control is empty and `INVALID` again (note the invalid styling on the field):

after - cleared value committed to the control

## Tests

The existing `'should handle updateOn blur strategy'` test was inert — it built a `FormGroup` that was never bound to the fixture and its only assertion was `expect(inputEl.nativeElement).toBeTruthy()`. It is replaced by a real bound host (`TestUpdateOnBlurPasswordComponent`) covering:

- typing does not update the control;
- blur commits the value, marks it touched/dirty and runs the validators;
- `minlength` errors surface once the value is committed;
- the clear icon commits the cleared value (the regression test for this fix — it fails without the one-line change).

The tests live in `password.test.ts` (Vitest) only — the Karma `password.spec.ts` changes were dropped after the rebase onto `main`, where Karma has been removed. `pnpm run test:unit` (7281 passing) and `pnpm run format:check` are both green. The clear-icon test fails without the one-line change in `password.ts` (`AssertionError: expected 'secret12' to be null`).

---

Co-authored by Claude.

### DIFF
--- a/packages/optimus-ui/src/password/password.test.ts
+++ b/packages/optimus-ui/src/password/password.test.ts
@@ -96,6 +96,24 @@ class TestFormPasswordComponent {
     toggleMask: boolean = true;
 }
 
+@Component({
+    changeDetection: ChangeDetectionStrategy.Eager,
+    standalone: false,
+    template: `
+        <form [formGroup]="form">
+            <p-password formControlName="password" [feedback]="false" [showClear]="true"> </p-password>
+        </form>
+    `
+})
+class TestUpdateOnBlurPasswordComponent {
+    form = new FormGroup({
+        password: new FormControl('', {
+            validators: [Validators.required, Validators.minLength(8)],
+            updateOn: 'blur' as const
+        })
+    });
+}
+
 // Comprehensive template test component with all ContentChild projections
 // Password pTemplate component
 @Component({
@@ -263,7 +281,7 @@ describe('Password', () => {
     beforeEach(async () => {
         await TestBed.configureTestingModule({
             imports: [PasswordModule, FormsModule, ReactiveFormsModule, CommonModule, SharedModule],
-            declarations: [TestBasicPasswordComponent, TestFormPasswordComponent, TestPasswordPTemplateComponent, TestPasswordRefTemplateComponent, TestPTPasswordComponent],
+            declarations: [TestBasicPasswordComponent, TestFormPasswordComponent, TestUpdateOnBlurPasswordComponent, TestPasswordPTemplateComponent, TestPasswordRefTemplateComponent, TestPTPasswordComponent],
             providers: [provideZonelessChangeDetection()]
         }).compileComponents();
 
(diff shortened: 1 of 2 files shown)


geromegrignon (changes requested):
Please rollback `spec.ts` changes as we removed Karma now.

rene-schakmann:
Rebased onto `main` and dropped the `password.spec.ts` changes — the file no longer exists there since Karma was removed. The coverage now lives only in `password.test.ts` (Vitest): typing does not commit under `updateOn: 'blur'`, blur commits and marks the control touched/dirty, `minlength` surfaces once committed, and the clear icon commits the cleared value. That last one is the regression test — it fails without the change in `password.ts` with `AssertionError: expected 'secret12' to be null`.

`pnpm run test:unit` is green (7281 passing, 90 skipped), as is `pnpm run format:check`. The diff is now just `password.ts` + `password.test.ts`.

PR #1567, opened 2026-08-20
fix(scroller): initialize virtual scroll when the viewport becomes measurable

sybers:
## Description

Open a `p-select` with `[virtualScroll]="true"` and don't move the mouse: the panel comes up empty. Wiggle the mouse and the options appear!

The overlay is still `display: none` when the scroller gets created (that's `p-motion` hiding the panel before it animates it in), so `isVisible()` is false when `viewInit()` runs and it gives up early. `last` stays at 0, so the rendered slice is empty. The only thing that retries is `onAfterViewChecked`, which needs a change detection cycle, and showing the panel doesn't schedule one. A mousemove over the panel does, hence the "move the mouse and it works" behaviour.

The reason it looks flaky rather than broken: `onOverlayBeforeEnter` does wake the scroller, but only through the `scrollToIndex` call that's guarded by "is something selected". So reopening a select that already has a value is usually fine, and the case that actually breaks is the first open on a placeholder.

This one is unrecoverable under `provideZonelessChangeDetection()`. No mousemove, no change detection, panel stays empty for good.

Fix is in two parts:

- `Scroller.viewInit()` now sets up a `ResizeObserver` when the element has no box yet, and initializes as soon as it gets one. Disconnects on init and on destroy. I left `onAfterViewChecked` in place because it's what arms the observer when the scroller starts out disabled and there's no element to observe yet.
- `p-select` and `p-multiselect` also call `scroller.viewInit()` from `onOverlayBeforeEnter`, where the panel is already laid out. `p-autocomplete` was already doing this, so it just brings the other two in line.

The scroller change also covers `p-listbox`, `p-table`, `p-treetable` and `p-tree`, which have no overlay hook and were relying on the same retry.

## Related issues

None found

## Type of change

- [ ] Documentation only

## Breaking changes

None, it is just a bugfix.

## Test plan

- [x] `pnpm run build:lib`
- [x] `pnpm run test:vitest` and `pnpm run test:unit`, both green
- [x] `pnpm run format:check`
- [x] `pnpm run lint`
- [ ] Demo app: didn't go through it. The new tests run in a real browser (Chromium for vitest, Chrome for karma) and cover the unmeasurable-viewport case directly.

Added 7 tests, mirrored in both the `.spec.ts` and `.test.ts` files so the jasmine and vitest jobs both pick them up. The scroller ones is in a new `Deferred Visibility Initialization` block.

## Checklist

- [ ] Issue discussed or bug clearly described (link issue when applicable)
- [x] Documentation updated (README, JSDoc, migration notes as needed)
- [x] Public API changes documented; breaking changes called out
- [x] CHANGELOG updated (if the repository maintains one and the change is user-facing)

### DIFF
--- a/packages/optimus-ui/src/multiselect/multiselect.spec.ts
+++ b/packages/optimus-ui/src/multiselect/multiselect.spec.ts
@@ -5,10 +5,10 @@ import { FormControl, FormGroup, FormsModule, NgForm, NgModel, ReactiveFormsModu
 import { By } from '@angular/platform-browser';
 import { provideNoopAnimations } from '@angular/platform-browser/animations';
 import { provideOptimus } from '@openng/optimus-ui/config';
+import type { MultiSelectBlurEvent, MultiSelectChangeEvent, MultiSelectFilterEvent, MultiSelectFocusEvent } from '@openng/optimus-ui/types/multiselect';
 import { BehaviorSubject, of } from 'rxjs';
 import { delay } from 'rxjs/operators';
 import { MultiSelect, MultiSelectModule } from './multiselect';
-import type { MultiSelectBlurEvent, MultiSelectChangeEvent, MultiSelectFilterEvent, MultiSelectFocusEvent } from '@openng/optimus-ui/types/multiselect';
 interface City {
     name: string;
     code: string;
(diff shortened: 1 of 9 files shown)


sybers:
Hey, any updates on this ? Could you take a look at the issue and suggested solution ?
Thanks 🙏

PR #1454, opened 2026-08-08
fix(cascadeselect): give the option list an accessible name

jase88:
## Description

The CascadeSelect option list bound `listlabel` instead of the `listLabel` getter, so the `aria-label` attribute was dropped and the list had no accessible name.

Before: `<ul role="tree">`
After: `<ul role="tree" aria-label="Option List">`

## Related issues

Fixes #1453

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that changes the public API)
- [ ] Documentation only
- [ ] Refactor, test, or chore (no user-facing change)

## Breaking changes

None

## Test plan

- [x] `ng test optimus-ui --watch=false --browsers=ChromeHeadless --include='**/cascadeselect.spec.ts'` — 111 passing
- [x] Confirmed the new test fails without the fix (`Expected null to be 'Option List'`)
- [x] `prettier --check` on both changed files
- [ ] Full `build` / `lint` not run locally

## Checklist

- [x] Issue discussed or bug clearly described (link issue when applicable)
- [x] Tests added or updated for behavioral changes
- [ ] Documentation updated (README, JSDoc, migration notes as needed)
- [ ] Public API changes documented; breaking changes called out
- [ ] CHANGELOG updated (if the repository maintains one and the change is user-facing)
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)
- [x] I agree to follow the [OpenNG Foundation Code of Conduct](https://github.com/openng-foundation/.github/blob/main/CODE_OF_CONDUCT.md)

## Additional context

Select, MultiSelect and AutoComplete already bind `listLabel` correctly.

PR #1568, opened 2026-08-20
fix(cascadeselect): bind role="group" on nested option lists

### LINKED ISSUE #1553: [Bug]: CascadeSelect: nested option lists render without role="group" ([attrrole] template typo)
### Checklist

### What happened?

While reviewing PR #1454 (the `listlabel` → `listLabel` aria-label fix) I spotted a second template typo of the same class a few lines above it, in `CascadeSelectSub`'s template.

The nested option list rendered for an expanded option group binds:

```html
<ul
    pCascadeSelectSub
    *ngIf="isOptionGroup(processedOption) && isOptionActive(processedOption)"
    [attrrole]="'group'"
    ...
></ul>
```

`[attrrole]` should be `[attr.role]`. As written it is neither an attribute binding nor an input of `pCascadeSelectSub`, so it does nothing: nested `<ul>` elements inside the `role="tree"` root list render without any `role`, breaking the tree/group semantics the template clearly intends for assistive technology.

The typo compiles silently because `packages/optimus-ui` builds with `strictTemplates: false` — the same enabler as the `listlabel` bug fixed by #1454.

### Steps to reproduce

1. Render a `<p-cascadeselect>` with grouped options (e.g. the countries-by-continent example from the docs).
2. Open the overlay and activate an option group so its nested list is displayed.
3. Inspect the nested `<ul>` rendered for the group's children.

### Expected behavior

The nested list carries `role="group"`, per the WAI-ARIA tree pattern: the root `<ul>` is `role="tree"` with `aria-orientation="horizontal"`, and descendant option lists should be group nodes.

### Actual behavior

The nested `<ul>` renders with no `role` attribute at all — the `[attrrole]` binding is silently dropped.

### Minimal reproduction

https://github.com/openng-org/optimus-ui/blob/f3376cf7250a5d5a71570f1ac7ec4debe0ee18cb/packages/optimus-ui/src/cascadeselect/cascadeselect.ts#L100

### Library version

2.0.0 (also present on current main)

### Angular version

22.0.8

### Node version

24.x (not version-dependent)

### Package manager

pnpm

### Browser(s)

Chrome

### Environment details

```markdown
Not environment-specific: the binding is dropped at template-compilation level, so the role attribute is missing in every browser/OS.
```

### Angular application setup

None

### Angular context

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional context

Found during review of PR #1454 — same bug class, same file. The fix is two characters: `[attrrole]` → `[attr.role]` in `packages/optimus-ui/src/cascadeselect/cascadeselect.ts` (line 100 on current main). Longer term, enabling `strictTemplates` for `packages/optimus-ui` would catch this whole class of template typo at compile time.

jase88:
Closes #1553

## What

The nested `<ul>` rendered for an expanded option group in `CascadeSelectSub` bound `[attrrole]="'group'"` instead of `[attr.role]="'group'"`. That is neither an attribute binding nor an input of `pCascadeSelectSub`, so the binding was silently dropped and nested lists rendered with no `role` at all inside the `role="tree"` root list — breaking the WAI-ARIA tree/group semantics the template intends.

Same bug class as the `listlabel` → `listLabel` fix in #1454, in the same file.

## Changes

- `packages/optimus-ui/src/cascadeselect/cascadeselect.ts`: `[attrrole]` → `[attr.role]`.
- `packages/optimus-ui/src/cascadeselect/cascadeselect.test.ts`: regression test that opens the overlay, activates an option group, and asserts the nested `<ul>` carries `role="group"`.

## Verification

Ran `ng run optimus-ui:test-vitest --include='**/cascadeselect/*.test.ts'`: 111/111 pass. With the typo restored the new test fails with `NG0303: Can't bind to 'attrrole' since it isn't a known property of 'ul'`, so it genuinely covers the regression.

Test added to `cascadeselect.test.ts` (vitest) only, not the identical Karma `cascadeselect.spec.ts`.

## Note

This compiled silently because `strictTemplates` and `fullTemplateTypeCheck` are `false` in `tsconfig.vitest.json`. Enabling `strictTemplates` for `packages/optimus-ui` would catch this whole class of template typo at build time — out of scope here, but worth a follow-up.

https://claude.ai/code/session_011xBTEbSbBHZJ8zXFw5YgsL

### DIFF
--- a/packages/optimus-ui/src/cascadeselect/cascadeselect.test.ts
+++ b/packages/optimus-ui/src/cascadeselect/cascadeselect.test.ts
@@ -854,6 +854,32 @@ describe('CascadeSelect', () => {
             expect(ariaRequired === null || ariaRequired === 'false').toBe(true);
             expect(hiddenInput.nativeElement.getAttribute('aria-label')).toBeTruthy();
         });
+
+        it('should render nested option lists with role="group"', async () => {
+            testComponent.options = mockCountries;
+            testFixture.changeDetectorRef.markForCheck();
+            await testFixture.whenStable();
+            testFixture.detectChanges();
+
+            const trigger = testFixture.debugElement.query(By.css('.p-cascadeselect-dropdown'));
+            trigger.nativeElement.dispatchEvent(new MouseEvent('mousedown', { bubbles: true }));
+            testFixture.changeDetectorRef.markForCheck();
+            testFixture.detectChanges();
+            await testFixture.whenStable();
+
+            const rootList = testFixture.debugElement.query(By.css('ul[role="tree"]'));
+            expect(rootList).toBeTruthy();
+
+            const groupOption = testFixture.debugElement.query(By.css('li[role="treeitem"]'));
+            groupOption.nativeElement.querySelector('.p-cascadeselect-option-content').click();
+            testFixture.changeDetectorRef.markForCheck();
+            testFixture.detectChanges();
+            await testFixture.whenStable();
+
+            const nestedList = rootList.nativeElement.querySelector('ul');
+            expect(nestedList).toBeTruthy();
+            expect(nestedList.getAttribute('role')).toBe('group');
+        });
     });
 
     describe('Complex Situations and Edge Cases', () => {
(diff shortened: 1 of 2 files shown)


geromegrignon:
Thanks!

PR #1579, opened 2026-08-23
fix(select): keep the lazy virtual scroll list at its full height

yelhouti:
## Problem

With a lazy virtual-scrolled `p-select` (`[virtualScroll]="true" [lazy]="true"`), the overlay list collapses to the height of an empty list, and the component can throw before the first page of options arrives.

Two separate causes:

**1. `Scroller.calculateAutoSize()` measures before the lazy items render.**

`autoSize` (which `p-select` and `p-multiselect` always turn on for their scrollers) measures the content element inside `Promise.resolve().then(...)`. That is a microtask, so it runs before Angular has rendered the items that the `onLazyLoad` handler just supplied — the content element is still empty, and the scroller pins itself to that empty height. With non-lazy options the items are already in the DOM when the measurement runs, so the bug only shows up when data arrives asynchronously.

Switching to a macrotask (`setTimeout(..., 1)`) lets the pending render flush first, so the measurement sees the real content.

**2. `Select.isOptionGroup()` throws on a not-yet-loaded option.**

A lazy data source typically hands back a sparse array sized to the total record count, with only the requested window filled in. The template calls `isOptionGroup(option)` for every rendered row, and for a hole in that array the current `option.optionGroup` access throws `Cannot read properties of undefined`. `p-multiselect` and the rest of the option helpers already guard against this; `Select.isOptionGroup` is the one that does not.

## Changes

- `scroller`: `calculateAutoSize()` defers its measurement to a macrotask, with a comment explaining why the microtask is not enough.
- `select`: `isOptionGroup()` uses optional chaining, matching how the other option helpers handle undefined entries.

## Tests

Added a `Select Lazy Virtual Scroll` suite to both `select.spec.ts` (jasmine) and `select.test.ts` (vitest):

- renders the lazy-loaded options and asserts the scroller keeps its full height — this fails on `main` because the scroller collapses;
- asserts `isOptionGroup(undefined)` does not throw when `optionGroupLabel` is set.

### DIFF
--- a/packages/optimus-ui/src/scroller/scroller.ts
+++ b/packages/optimus-ui/src/scroller/scroller.ts
@@ -900,7 +900,9 @@ export class Scroller extends BaseComponent<VirtualScrollerPassThrough> {
 
     calculateAutoSize() {
         if (this._autoSize && !this.d_loading) {
-            Promise.resolve().then(() => {
+            // a macrotask is required here: when items are lazy loaded, the new items are only
+            // rendered on the next tick, and a microtask would measure the still empty content
+            setTimeout(() => {
                 if (this.contentEl) {
                     this.contentEl.style.minHeight = this.contentEl.style.minWidth = 'auto';
                     this.contentEl.style.position = 'relative';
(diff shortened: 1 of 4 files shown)


geromegrignon (changes requested):
Review is fine, please delete the `.spec.ts` file changes as we dropped Karma and spec files lately.

PR #1604, opened 2026-08-26
fix(datepicker): type the custom date format

### LINKED ISSUE #739: Component: Calendar missing Date Type
### Describe the bug

The Calendar component has a special date type that is not defined. It differs strongly from the native Date type, it appears to be used primarily for display rather than calculation

### Environment

Typescript:  5.4.5
Angular: 18.2.8

### Reproducer

_No response_

### Angular version

18.2.8

### PrimeNG version

17.18.12

### Build / Runtime

Angular CLI App

### Language

TypeScript

22.5.1

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

> Migrated from `primefaces/primeng#16835` — originally by `@zachary-svoboda-accesso` on 2024-11-25

jase88:
Closes #739

`Month.dates` was typed `Date[]` but actually holds weeks of day metadata. Added `DatePickerDay` / `DatePickerWeek` and replaced the `any` dateMeta params with them.

Port of primefaces/primeng#17273. Credits to @zachary-svoboda-accesso.

### DIFF
--- a/packages/optimus-ui/src/datepicker/datepicker.ts
+++ b/packages/optimus-ui/src/datepicker/datepicker.ts
@@ -40,6 +40,7 @@ import { Nullable, VoidListener } from '@openng/optimus-ui/ts-helpers';
 import {
     DatePickerButtonBarTemplateContext,
     DatePickerDateTemplateContext,
+    DatePickerDay,
     DatePickerDecadeTemplateContext,
     DatePickerDisabledDateTemplateContext,
     DatePickerInputIconTemplateContext,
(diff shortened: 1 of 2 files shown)


packages/optimus-ui/src/datepicker/datepicker.ts
```diff
@@ -1551,12 +1553,7 @@ export class DatePicker extends BaseInput<DatePickerPassThrough> {
             (dates as any[]).push(week);
```
geromegrignon: could you remove the cast as not useful anymore?

packages/optimus-ui/src/datepicker/datepicker.ts
geromegrignon: Remove the cast here too

geromegrignon (changes requested):

PR #1610, opened 2026-08-27
fix(schematics): fail ng-add on a PrimeNG workspace instead of warning

### LINKED ISSUE #1447: [Bug]: ng add on a PrimeNG workspace prompts for a theme, installs the package, then silently does nothing
### Checklist

### What happened?

Split out of #1444, which was about the docs being wrong. This one is about the behaviour itself being hard to read.

Running `ng add @openng/optimus-ui` in a PrimeNG v21 workspace asks a setup question, installs the package, makes no changes, and exits 0. The sequence reads as "this worked".

In order:

1. The CLI installs `@openng/optimus-ui` and writes it into `dependencies`. `packages/optimus-ui/package.json` declares `"ng-add": { "save": "dependencies" }`, and the CLI only removes the temporary dependency when `save` is `false`.
2. The CLI resolves the schematic's `x-prompt`s before invoking the factory, so the user is asked **"Which theme preset would you like to use?"** and picks one.
3. `ngAdd` runs, hits the `hasPrimeng` guard at `schematics/ng-add/index.ts:38`, logs a warning and returns the tree unchanged. The theme answer is discarded.
4. Because it is `logger.warn` and not a throw, the command exits 0.

### Steps to reproduce

1. Create an Angular v21 workspace with `primeng` in `dependencies`.
2. Run `ng add @openng/optimus-ui`.
3. Answer the "Which theme preset would you like to use?" prompt.
4. Inspect `package.json`, the lockfile, and the exit code.

### Expected behavior

Either the command declines clearly (non-zero exit, no dependency left behind), or it does not ask a setup question whose answer it will discard.

### Actual behavior

The user is left with `primeng` and an unwired `@openng/optimus-ui` both in `dependencies`, a modified lockfile, a discarded answer to a prompt, and a zero exit code. The warning is easy to miss in the middle of the output, and the new dependency in `package.json` looks like confirmation that something was set up.

### Minimal reproduction

No external repro needed, reproducible directly against this repository. See the steps above.

### Library version

1.0.1, 2.0.0-rc.0, and `main` at 820a69ad89

### Angular version

21.0.0

### Node version

24.17.0

### Package manager

pnpm

### Browser(s)

Other

### Environment details

- OS: macOS
- Node: 24.17.0

### Angular application setup

Both / mixed

### Relevant log output

```shell
primeng detected — ng-add only sets up Optimus UI in new projects, so no changes were made.
To migrate this workspace, run: ng generate @openng/optimus-ui:migrate-from-primeng
```

### Additional context

**Possible directions**, not sure which is preferable, hence the issue rather than a PR:

- Throw a `SchematicsException` instead of warning, so the CLI reports failure and rolls the install back. Clearest signal, but `ng add` failing on a supported workspace may be worse UX than a warning.
- Keep the warning but suppress the theme prompt when `primeng` is present, so the user is not asked a question whose answer is thrown away.
- Offer the migration as a prompt: "primeng detected, run the migration instead?" and delegate to `migrate-from-primeng` on yes. This is close to the pre-`1.0.0-rc.2` behaviour, which was deliberately removed in 851df785bf, so it may well be a non-starter, listing it for completeness.

At minimum, the theme prompt firing before a no-op bail seems worth fixing regardless of which route is taken.

**Side note.** The `hasPrimeng` bail at `schematics/ng-add/index.ts:38` runs before the `--project` validation at `:91-94`. So `ng add @openng/optimus-ui --project=does-not-exist` in a PrimeNG workspace warns and exits 0 instead of erroring on the bad project name.

**Affected versions.** `schematics/ng-add/index.js` is byte-identical across `1.0.0-rc.2`, `1.0.0`, `1.0.1` (`latest`) and `2.0.0-rc.0` (`rc`).

rene-schakmann:
Closes #1447.

## The problem

`ng add @openng/optimus-ui` on a PrimeNG v21 workspace asked for a theme preset, installed the package, changed nothing, and exited 0. Every signal the user got said "this worked": a setup question answered, a new entry in `dependencies`, a modified lockfile, and a `logger.warn` in the middle of the output.

The bail itself is correct — `ng add` only sets up fresh projects, migration is `migrate-from-primeng`'s job — but a warning is the wrong way to report it.

## The change

`schematics/ng-add/index.ts` throws a `SchematicsException` instead of warning, so the CLI reports the refusal as a failure with a non-zero exit code. Nothing in the workspace is touched either way: a failed schematic never commits its tree.

Of the three directions listed in the issue, this is the first one. The second (suppress the theme prompt when primeng is present) is not implementable — as [@geromegrignon noted](https://github.com/openng-org/optimus-ui/issues/1447#issuecomment-5218327220), the CLI resolves `x-prompt`s before the factory runs, so the schematic cannot influence them. The third (offer to run the migration) was deliberately removed in 851df785bf. Failing loudly is what is left, and it matches the first branch of the issue's expected behaviour.

One caveat worth stating plainly, because the issue assumed otherwise: **the CLI does not roll the install back.** `AddCommandModule` only removes the dependency it added when `ng-add.save` is `false` (`node_modules/@angular/cli/src/commands/add/cli.js`, the `shouldCleanUp` branch), and ours is `"dependencies"`. So `@openng/optimus-ui` stays in `package.json` after the failure. That is not purely a wart — `ng generate @openng/optimus-ui@1:migrate-from-primeng` needs the package installed to resolve — so the new message says so instead of leaving the user to guess:

```
primeng detected — ng add only sets up Optimus UI in new projects, so no changes were made and the theme preset was not configured.
To migrate this workspace, run: ng generate @openng/optimus-ui@1:migrate-from-primeng
The Angular CLI has already added @openng/optimus-ui to your dependencies. The migration schematic needs it there, so keep it — or remove it with your package manager if you did not mean to add Optimus UI to this workspace.
```

Two smaller things folded in:

- **The side note from the issue.** An explicit `--project` is now validated *before* the primeng check, so `ng add @openng/optimus-ui --project=does-not-exist` on a PrimeNG workspace reports the typo rather than having it masked by whichever check ran first.
- **The migration hint is now pinned to `@1`.** `migrate-from-primeng` is only maintained on `release/1.x` (per its own `@deprecated` tag and the README), so the unpinned `ng generate @openng/optimus-ui:migrate-from-primeng` in both ng-add messages pointed at the wrong line.

Docs updated to match in `README.md`, `apps/docs/doc/installation/ngadd-doc.ts` and `apps/docs/doc/faq/migration-doc.ts` — all three said ng-add "makes no changes" when primeng is found, which now undersells it. The generated files under `apps/docs/public/llms` are left to their build.

## Tests

`packages/optimus-ui/schematics/test/ng-add.spec.ts` — the eight existing primeng cases now assert rejection instead of a warning, plus three new ones:

- the failure is a `SchematicsException` and the workspace files are byte-identical afterwards
- the message names the state the CLI left behind (`already added @openng/optimus-ui to your dependencies`, the `@1`-pinned migrate command)
- a mistyped `--project` in a PrimeNG workspace reports the bad project name, not the primeng bail

All 11 fail on `main` and pass here; the full schematics suite is 134 passing.

## Proof of work

The compiled schematic driven through `NodeWorkflow` — the same workflow `ng add` uses — against a real on-disk PrimeNG v21 workspace, before and after, plus the fresh-workspace path to show it is unaffected:

ng add on a PrimeNG workspace: before/after terminal output and the passing schematics suite

---

Co-authored by Claude.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Full setup, including the manual path and icons, is in the [getting started guid
 
 ## Migrating from PrimeNG
 
-`ng add` does not migrate an existing workspace. It sets up new projects only, and when it detects a `primeng` dependency it leaves your code untouched and points you here. Use the `migrate-from-primeng` schematic instead: install the package so the Angular CLI can resolve it, then run the schematic. Packages are swapped, imports are rewritten, and you get a report of anything it could not handle automatically, with file and line numbers.
+`ng add` does not migrate an existing workspace. It sets up new projects only, and when it detects a `primeng` dependency it leaves your code untouched and fails with an error pointing you here rather than reporting success. Use the `migrate-from-primeng` schematic instead: install the package so the Angular CLI can resolve it, then run the schematic. Packages are swapped, imports are rewritten, and you get a report of anything it could not handle automatically, with file and line numbers.
 
 ```bash
 npm install @openng/optimus-ui@1
--- a/apps/docs/doc/faq/migration-doc.ts
+++ b/apps/docs/doc/faq/migration-doc.ts
@@ -16,7 +16,7 @@ import { RouterModule } from '@angular/router';
 
             <h3>How do I migrate from PrimeNG? Is it automatic?</h3>
             <p>
-                Mostly, but not through <i>ng add</i>, which only sets up new projects and makes no changes when it finds an existing <i>primeng</i> dependency. Install <i>&#64;openng/optimus-ui</i>, then run
+                Mostly, but not through <i>ng add</i>, which only sets up new projects and fails without changing anything when it finds an existing <i>primeng</i> dependency. Install <i>&#64;openng/optimus-ui</i>, then run
                 <i>ng generate &#64;openng/optimus-ui:migrate-from-primeng</i> in a PrimeNG v21 workspace: packages swapped, imports rewritten, dependencies installed. It then prints a report of anything it could not rewrite, with file and line
                 numbers. The <a href="https://v1.optimus.openng.org/migration/primeng" target="_blank" rel="noopener noreferrer" class="doc-link">migration guide</a> covers the details and the manual cases.
             </p>
--- a/apps/docs/doc/installation/ngadd-doc.ts
+++ b/apps/docs/doc/installation/ngadd-doc.ts
@@ -20,9 +20,10 @@ import { Component } from '@angular/core';
             <p>Pass the preset up front to skip the prompt, or skip the install step if you want to run it yourself:</p>
             <app-code [code]="flagsCode" [hideToggleCode]="true"></app-code>
             <p>
-                If the schematic cannot find a providers array to update it prints the three manual steps instead of guessing, so nothing in your workspace is rewritten unexpectedly. If it detects an existing <i>primeng</i> dependency it makes no
-                changes to your code and points you at the <a href="https://v1.optimus.openng.org/migration/primeng" target="_blank" rel="noopener noreferrer" class="doc-link">migration guide</a>, which covers the
-                <i>migrate-from-primeng</i> schematic. It does not run that schematic for you.
+                If the schematic cannot find a providers array to update it prints the three manual steps instead of guessing, so nothing in your workspace is rewritten unexpectedly. If it detects an existing <i>primeng</i> dependency it changes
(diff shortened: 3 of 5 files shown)


geromegrignon:
Thanks @rene-schakmann !

PR #1612, opened 2026-08-27
fix: mark library-internal ngModel bindings as standalone (NG01354)

rgroenewoudt:
## Problem

Angular **22.1.3** added a dev-mode warning, **NG01354**:

> `ngModel` on a form control inside a child component cannot register with the `FormGroupDirective` in the parent component because `@Host()` stops injection at the component boundary.

It fires from `NgModel.ngOnChanges` when the NgModel has no `@Host()` parent, is not marked standalone, and a plain element-injector lookup can still reach a `ControlContainer`:

```ts
if (!this._registered && ngDevMode && this._parent === null && !this.options?.standalone) {
    const parentContainer = this._ngModelInjector?.get(ControlContainer, null);
    if (parentContainer != null) console.warn(ngModelInChildComponentWarning(typeName));
}
```

The library binds `[ngModel]` on components it renders internally (Checkbox, ToggleButton, Select, InputNumber, RadioButton) in 23 places across 10 components. As soon as an application puts one of those components inside a `[formGroup]` or a `<form>`, the console fills with warnings it cannot act on — the binding lives inside the library's own template.

**Reproduction:** put any of `p-selectbutton`, `p-multiselect`, `p-listbox`, `p-paginator`, `p-table`, `p-tree`, `p-treetable`, `p-picklist`, `p-orderlist`, `p-terminal` inside a `[formGroup]` on Angular ≥ 22.1.3 and open the dev console. (For `p-multiselect` the panel has to be opened — its bindings live in the overlay.)

## Fix

Mark every one of those bindings `[ngModelOptions]="{ standalone: true }"`.

## Why this is the right fix, and why it is safe

- **These NgModels never registered with a consumer form anyway.** `NgModel` injects its parent as `{ token: ControlContainer, host: true, optional: true }`, and `@Host()` stops at the component view boundary — `_parent` is `null` for all 23. The warning reports an existing condition; it does not describe a new one.
- **`standalone: true` is behaviourally inert here.** It is read in `_isStandalone()` (`!this._parent || options.standalone` — already `true` via the first clause) and in the warning guard above. With `_parent === null`, the flag only takes these bindings out of the warning's scope.
- **The alternative would break consumers.** Exposing the container via `viewProviders: [{ provide: ControlContainer, useExisting: … }]` — the other route the warning suggests — makes `_checkForErrors()` run: a reactive parent then throws **NG01350** (`ngModel` cannot register with a parent `formGroup`) and a template-driven parent **NG01352** (`missingNameException`), since none of these bindings carry an NgModel name. Where one appears to (`p-tableCheckbox`'s `[name]`), that is `BaseEditableHolder.name` and it is identical for every row.
- **Nothing in the library ever opted in:** `grep -rn "ControlContainer" packages/optimus-ui/src` returns no hits. Every one of these bindings carries derived rendering state (`isSelected(option)`, `allSelected()`, `getPage()`, `checked`, `filterConstraint?.value`), while the component's real value already reaches the parent form through its own `ControlValueAccessor`.

## Tests

Added a regression test to `selectbutton`, `multiselect`, `listbox`, `paginator` and `tree` that renders the component inside a `[formGroup]` and asserts that every `NgModel` in its view is standalone, that no `NG01354` warning is logged, and that the parent form's controls are untouched. All five fail on `main` and pass with this change.

Both assertions are live: `main` now resolves `@angular/forms` **22.1.4**, so the warning is really emitted. The `NgModel`-is-standalone assertion is kept alongside it because it states the intent directly and does not depend on the wording or the error code of a dev-mode message.

## Verification

- `pnpm --filter @openng/optimus-ui test:unit` — **7276 passed, 90 skipped, 0 failed**, rebased on `main` (3d5438e).
- `pnpm --filter @openng/optimus-ui build` — clean.
- `prettier --check` on all changed files — clean. The added attribute pushes six template lines past `printWidth: 250`, so Prettier reflows those three elements (`paginator`, `table`, `treetable`) across several lines; that formatting is included here rather than left for a later PR to pick up.

Two pre-existing issues on `main`, unrelated to this change and left untouched: `pnpm run lint` fails immediately (`--ignore-path` is no longer accepted alongside `eslint.config.js`), and `pnpm run test:unit`'s `build:optimus-ui` step matches no projects on Windows (`pnpm --filter './packages/optimus-ui-*'`), so the suite was run through the package filter directly.

## Follow-up (not in this PR)

The deeper cleanup is to stop using `ngModel` internally at all — every one of these is a one-way "render as checked/selected" binding, so a plain state input on Checkbox/ToggleButton would express it without a form directive. That is a public API addition across several components and belongs in its own PR.


rgroenewoudt:
If I run prettier I get additional whitespace changes that should be on main already.

geromegrignon:
@rgroenewoudt prettier format the template due to the change.
Please run it to avoid having such a change introduced by another PR affecting these files while not changing this part of the template.

geromegrignon (changes requested):
see above comment

rgroenewoudt:
Prettier format done.

geromegrignon (approved):

geromegrignon:
Thanks @rgroenewoudt , merged and released with **2.0.2**

PR #1614, opened 2026-08-27
fix(dynamicdialog): respect ariaLabelledBy from config

### LINKED ISSUE #423: DynamicDialog: ariaLabelledBy from DynamicDialogConfig not considered
### Describe the bug

`DynamicDialogConfig` allows passing in a property `ariaLabelledBy`.
But this property is not considered.

The attribute `aria-labelledby` is set to the header title element (if available), but sometimes it is necessary to link it to a different element.

### Pull Request Link
https://github.com/primefaces/primeng/pull/1136

### Reproducer

https://stackblitz.com/github/primefaces/primeng-issue-template

### Environment

MacOS

### Angular version

19.2.1

### PrimeNG version

v19

### Node version

_No response_

### Browser(s)

Chrome

1. Call the method `open` on `DynamicDialog` service with the configuration `{ariaLabelledBy: 'notWorking'}`
2. Look at the DO

### Expected behavior

The ariaLabelledBy property from the config object should have a higher priority than the automatic reference to the dialog title.

> Migrated from `primefaces/primeng#18488` — originally by `@jase88` on 2025-06-16

jase88:
Fixes #423

`DynamicDialogConfig.ariaLabelledBy` was ignored. Now it wins over the generated header id.

Port of primefaces/primeng#18489, plus 2 tests.

### DIFF
--- a/packages/optimus-ui/src/dialog/dialog.test.ts
+++ b/packages/optimus-ui/src/dialog/dialog.test.ts
@@ -45,6 +45,7 @@ import { Dialog } from './dialog';
             [closeButtonProps]="closeButtonProps"
             [maximizeButtonProps]="maximizeButtonProps"
             [role]="role"
+            [ariaLabelledBy]="ariaLabelledBy"
             (onShow)="onShowEvent($event)"
             (onHide)="onHideEvent($event)"
             (onMaximize)="onMaximizeEvent($event)"
@@ -92,6 +93,7 @@ class TestBasicDialogComponent {
     closeButtonProps: any = {};
     maximizeButtonProps: any = {};
     role = 'dialog';
+    ariaLabelledBy: string | undefined;
     breakpoints: any = null as any;
 
     // Event handlers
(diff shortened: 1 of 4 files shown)


geromegrignon:
Hi @jase88, it would need an extra fix as `DynamicDialog.ariaLabelledBy` is never used.
We now pass the configuration but no one use it.

The `ariaLabelledBy` is needed for the child Dialog component.

So it would need:
- to create an input for the `Dialog` component to grab it
- to update the `Dialog` logic to resolve the `ariaLabelledBy` through the input or its existing `getAriaLabelledBy`
- to update the `Dialog`template accordingly if a property is renamed
- to bind it in the `DynamicDialog` template

geromegrignon (changes requested):
See above comment

jase88:
> See above comment

Thanks! Pushed                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                          
  - `Dialog`: new `ariaLabelledBy` signal input; `header` is a signal input too, `aria-labelledby` is a `computed()` from both.                                                                                                                                                                                           
  - `DynamicDialog`: binds `[ariaLabelledBy]="ddconfig?.ariaLabelledBy"`.                                                                                                                                                                                                                                                 
  - Tests added for default id, override, and config → `Dialog` binding.                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                          
  Note: `header` is now a signal, so `dialog.header` → `dialog.header()` for programmatic readers (none in repo). Can revert that part if you'd rather keep this PR minimal.

geromegrignon (approved):

PR #1757, opened 2026-09-03
fix(chart): declare chart.js as an optional peer dependency

rgroenewoudt:
## Problem

`packages/optimus-ui/src/chart/chart.ts` does `import Chart from 'chart.js/auto'`, but `@openng/optimus-ui` declares `chart.js` neither as a dependency nor as a peer dependency. The published `package.json` lists only the Angular packages and `rxjs` as peers.

**Reproduction:** an Angular 22 app that imports `ChartModule` from `@openng/optimus-ui/chart`, installed with pnpm 11 and `enableGlobalVirtualStore: true`. `chart.js` is present in the app's `package.json`; esbuild resolves the import relative to the library file in the store and never finds it. pnpm's resolution is more strict and only allows resolving libraries if set in package.json.

## Fix
Declare `chart.js` as an **optional** peer dependency (`^4.0.0`, matching the `4.5.1` used by `apps/docs`) via `peerDependencies` + `peerDependenciesMeta`.

- Optional, so consumers that never use `p-chart` are not forced to install it and get no missing-peer warning.
- Only `package.json` and the lockfile change

## Summary by CodeRabbit

* **New Features**
  * Added optional Chart.js support for Optimus UI integrations.

### DIFF
--- a/packages/optimus-ui/package.json
+++ b/packages/optimus-ui/package.json
@@ -77,6 +77,7 @@
         "@angular/forms": "catalog:angular22",
         "@angular/platform-browser": "catalog:angular22",
         "@angular/router": "catalog:angular22",
+        "chart.js": "^4.0.0",
         "rxjs": "^6.0.0 || ^7.8.1"
     },
     "dependencies": {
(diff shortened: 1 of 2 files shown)


SoyDiego (approved):

Ionaru (approved):

geromegrignon:
Thanks @rgroenewoudt (including @SoyDiego and @Ionaru for the review!)

PR #1784, opened 2026-09-08
fix(tooltip): wait for dialog motion before showing

devfake:
## Description

Tooltips inside a dialog currently wait a fixed 100 ms before becoming visible, even when the dialog is already open and stationary. The 250 ms fade starts while the tooltip is still hidden, so it appears late and partway through its animation compared with the same tooltip outside a dialog.

This change replaces the fixed timeout with a wait for active, finite animations on the actual ancestor dialog panels. Once the motion finishes, the tooltip is displayed, aligned and faded in. Tooltips in settled dialogs appear immediately, respecting the configured show delay.

This also preserves the original autofocus positioning fix: when an input receives focus during dialog entrance, its tooltip waits until the panel reaches its final position. Nested dialogs and dialogs appended to the body are supported. Animations on child elements do not delay the tooltip, and pending callbacks cannot reveal a dismissed or replaced tooltip.

## Related issues

Related historical issue: [primefaces/primeng#15485](https://github.com/primefaces/primeng/issues/15485).

Preserves the positioning fix introduced in [primefaces/primeng#15486](https://github.com/primefaces/primeng/pull/15486), replacing its fixed timeout with the actual dialog animation lifecycle.

## Type of change

- [ ] Documentation only

## Breaking changes

None

## Test plan

- [x] `npm run build`
- [x] `npm test`
- [ ] `npm run lint`
- [ ] Verified in the demo app (if applicable)

## Checklist

## Additional context

Before

https://github.com/user-attachments/assets/0e073a93-9e1e-4e52-ae8f-1486197c7701

After

https://github.com/user-attachments/assets/143e61c4-263a-4f76-a100-35958600bb69

The before recording uses the existing implementation, which already includes the historical autofocus fix.

The `ms` values show the elapsed time from the hover or focus event to the first observed animation frame where the tooltip is displayed with an opacity greater than zero. They measure the delay until the tooltip starts appearing, not the duration of the full fade-in. 

The videos were recorded in Chrome. In Firefox, the effect is worse, with the tooltip almost popping into view instead of fading in smoothly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tooltip timing while dialog surfaces are animating, including paused and nested animations.
  * Prevented tooltips from appearing in stale containers or after their associated interface elements are no longer connected.
  * Ensured tooltips reveal immediately when no dialog animation is active.
  * Improved tooltip positioning and fade-in behavior after the appropriate display timing is reached.

### DIFF
--- a/packages/optimus-ui/src/tooltip/tooltip.ts
+++ b/packages/optimus-ui/src/tooltip/tooltip.ts
@@ -579,7 +579,7 @@ export class Tooltip extends BaseComponent<TooltipPassThroughOptions> {
 
         const reveal = () => {
             // A hide, destroy or subsequent show invalidates this pending reveal.
-            if (this.container !== container || !nativeElement.isConnected) {
+            if (!this.active || this.container !== container || !nativeElement.isConnected) {
                 return;
             }
 
--- a/packages/optimus-ui/src/tooltip/tooltip.ts
+++ b/packages/optimus-ui/src/tooltip/tooltip.ts
@@ -566,20 +566,34 @@ export class Tooltip extends BaseComponent<TooltipPassThroughOptions> {
 
         this.create();
 
-        const nativeElement = this.el.nativeElement;
-        const pDialogWrapper = nativeElement.closest('p-dialog');
+        const nativeElement = this.el.nativeElement as HTMLElement;
+        const container = this.container;
+        const dialogSelector = '.p-dialog, [data-pc-name="dialog"]';
+        const dialogAnimations: Animation[] = [];
 
-        if (pDialogWrapper) {
-            setTimeout(() => {
-                this.container && (this.container.style.display = 'inline-block');
-                this.container && this.align();
-            }, 100);
-        } else {
-            this.container.style.display = 'inline-block';
-            this.align();
+        // Inspect dialog surfaces, not descendants: button hover/focus transitions
+        // must not delay a settled dialog's tooltip. This also works with appendTo="body".
+        for (let dialog = nativeElement.closest(dialogSelector); dialog; dialog = dialog.parentElement?.closest(dialogSelector) ?? null) {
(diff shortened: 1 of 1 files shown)


Ionaru (approved):

Ionaru:
@devfake The second commit you pushed failed to pass the lint check. Should be fine if you prefix it with `fix(tooltip):` and force-push.

packages/optimus-ui/src/tooltip/tooltip.ts
```diff
@@ -566,20 +566,34 @@ export class Tooltip extends BaseComponent<TooltipPassThroughOptions> {
...
-            }, 100);
-        } else {
-            this.container.style.display = 'inline-block';
-            this.align();
+        const nativeElement = this.el.nativeElement as HTMLElement;
+        const container = this.container;
+        const dialogSelector = '.p-dialog, [data-pc-name="dialog"]';
+        const dialogAnimations: Animation[] = [];
+
+        // Inspect dialog surfaces, not descendants: button hover/focus transitions
```
dominicbachmann: I think this comment can be removed

packages/optimus-ui/src/tooltip/tooltip.ts
```diff
@@ -566,20 +566,34 @@ export class Tooltip extends BaseComponent<TooltipPassThroughOptions> {
...
+        // must not delay a settled dialog's tooltip. This also works with appendTo="body".
+        for (let dialog = nativeElement.closest(dialogSelector); dialog; dialog = dialog.parentElement?.closest(dialogSelector) ?? null) {
```
dominicbachmann: I think this will be more readable with a while-loop

dominicbachmann:
I created an alternative pr https://github.com/openng-org/optimus-ui/pull/1804/changes that fixes the issue as well and also works with elements other than dialog

dominicbachmann (changes requested):
As mention in another comment please check https://github.com/openng-org/optimus-ui/pull/1804

PR #1790, opened 2026-09-09
feat(taskboard): add taskboard component

nesgarbo:
Adds a TaskBoard component: `p-taskboard-root` owns the cards, the layout state, the interaction and the outputs, and every child part reads what it needs from that shared state instead of receiving a dozen inputs from a parent that would have to forward them. Nothing visible ships in the structural parts — what they guarantee is the contract the visible content sits inside: focus, selection, drag geometry, workflow validation, data attributes and ARIA.

The implementation is original: the indexing engine, the pointer sensor, the keyboard controller, the workflow validation, the parts and the styles are written for this repo, and they follow the conventions the rest of optimus-ui already uses — signal state on a root, `*...Def` template definitions, `dt('component.*')` design tokens, a `BaseStyle` per component.

Deciding what a board of this kind has to cover came from public material: PrimeUI Pro's published TaskBoard documentation, its markdown endpoints and its demo pages, https://angular.primeuipro.dev/taskboard, which is where the API surface — the compound part set, the option names, the payload shapes, the data attributes — and the visual reference come from. Matching the visuals also meant reading the compiled stylesheet the public demo serves to the browser, so the spacings, the state layers and the responsive breakpoints are ports of observed values rather than guesses; the stylesheet here is written against optimus-ui's token engine, not copied. No PrimeUI Pro source, package or npm bundle was consulted.

Entry point is `@openng/optimus-ui/taskboard`; tokens live in the aura, lara and nora presets under `taskboard.*`.

## What is in it

**Layout** — columns from a `columns` input or from projected `p-taskboard-column` children; swimlanes as a second grouping dimension that scopes placement and move indexes to the CELL rather than the column; phase headers spanning runs of columns; collapse, reorder, locked and pinned columns.

**Phase headers re-split rather than stretch.** The header band walks the visible column order and opens a new segment whenever the group changes, so a group whose columns are no longer adjacent — because `order` moved one, or because `access.columnAccess` hid one — renders as two labelled runs instead of one label reaching across a lane it does not own. Their width is a `calc()` over the column-width token driven by a span count in the inline style, so the band follows every responsive change of the columns without measuring anything.

**Two data modes.** `[(tasks)]` lets the board apply accepted moves, reorders and mutations, emitting the next array BEFORE the matching domain output so a handler that reads the signal it just wrote already sees the change. `[items]` hands ownership to a store: the board never writes, `tasksChange` is not emitted at all, and the local history stays empty because undo belongs to whoever owns the data. Binding both is a development warning and not a thrown error — the board still renders, `items` wins, but it cannot guess which of the two was meant to own the writes.

**Only three fields are read.** The id named by `dataKey`, the column named by `columnField` and, on a grouped board, the row named by `swimlaneField`. Everything else on a record is the application's and travels untouched through the contexts and the payloads, so a board of tickets with `ticketId` and `stage` needs no mapping layer.

**Movement goes through one door.** The pointer, the keyboard and `moveTask` all call the same request path, so the three cannot end up validating differently: transition rules, then WIP capacity, then required fields, then confirmation. That order is not cosmetic — a card that is not allowed into a column at all should say so, not complain first about a field it was never going to be asked for. A refused move leaves the data untouched, emits `cardDropBlocked` with a reason and a message, and announces it in the assertive live region; a guarded one is held as a pending move until the projected `p-taskboard-drag-confirm` answers.

**The pointer sensor snapshots geometry.** Cells, cards and columns are measured when the drag starts and re-measured when something actually invalidates it — a scroll of the board or of a column. Hit-testing against live `getBoundingClientRect()` calls on every pointer move forces a layout per move and the drag visibly stutters. Insertion is decided by each card's MIDPOINT rather than its edges, because comparing against edges leaves a dead band between two cards where the marker does not move, which reads as the board ignoring the pointer. Edge auto-scroll clamps to the real bounds instead of running against a wall.

**The travelling preview is a clone of the card.** `<p-taskboard-drag-preview />` with no projected body shows a copy of the source card's element children, stripped of ids, tab stops and enabled controls — the runtime cannot know what an application's card renders, so it takes the rendered nodes. Cloned and never moved: the real card stays in its column, dimmed, until the drop is decided. Project content or a `pTaskBoardDragPreviewDef` to replace it.

**Markers reveal themselves.** An authored `p-taskboard-drop-indicator` compares its own index against the position the sensor publishes, so nothing writes `hidden` into the DOM behind Angular's back. A cell that authored none gets a single fixed runtime line instead, and projected content adds `p-taskboard-drop-indicator-custom` to suppress the preset line — the two appearances are mutually exclusive on purpose.

**Text selection cannot survive a drag.** `user-select: none` follows the PRESS rather than the drag, because the browser starts selecting the moment the pointer moves with the button down, well before the threshold; `selectstart` is cancelled while the press is tracked, since a card projects arbitrary content and an application stylesheet can put `user-select: text` back on part of it.

**Selection** — `single` and `multiple`, with modifier-toggle and Shift-range. A range is local to the cell: if the Shift target sits in another column or row, only that card is selected and the anchor moves there, because the order inside each column is independent. Dragging a selected card in multiple mode moves the whole group in its rendered order; dragging an unselected one moves just it and leaves the selection alone.

**Keyboard** — arrows across cards, columns, EMPTY columns and swimlane cells; Home/End; Tab between columns; Enter to activate; Space to toggle; Shift-arrows to extend a range; Ctrl/Cmd A per cell; Alt-arrows to reorder and to cross columns; Alt+Shift to cross swimlanes; Ctrl/Cmd Z/Y for the history. `Escape` unwinds one layer at a time, most transient first: pending confirmation, drag, column reorder, selection, focus. Board shortcuts step aside for anything that owns its own keys — native controls, editable regions, and elements carrying an interactive ARIA role — so `Space` on a button inside a card presses the button.

**Virtual scroll** — a per-cell window from the scroll offset, the measured card height and a buffer, with spacers reserving the rest. Off, or below one viewport's worth of cards, the window is the whole list: windowing a short column costs two spacer elements and buys nothing.

**Export, print and saved views** — `exportToJSON`, `exportToCSV` with field, delimiter and array-joiner options, the download helpers, and `serializeState`/`restoreState` for the UI state alone. `print()` marks the board as the print target and unclips its ancestors, which is the reason it exists at all: a board inside a scroll panel or a drawer is clipped by boxes it does not own, and `window.print()` on its own comes out as the one visible screenful. On paper the columns wrap instead of scrolling, the column body keeps its flex gap so the cards do not print touching each other, the collapse controls and count chips go, a tag keeps only its outline, and a collapsed column prints its cards — a saved screen state should not remove work from the printed board.

**Supplied visuals** — a card, a denser card, a column header with the WIP badge and the collapse chevron, and the two swimlane headers. They are the only components in the package that decide how anything looks, and they read their data from the nearest context, so `<p-taskboard-card-ui />` with no bindings is a complete card. A product that outgrows them drops its own component into the same wrapper and loses nothing.

## Contract

The board never mutates the array it was given. In managed mode it emits a REPLACEMENT array and the application writes it; in external mode it emits requests. `cardCreate`, `cardUpdate` and `cardDelete` are the same in both, and an output handler's return value cancels nothing.

Every part carries `data-scope="taskboard"` and a `data-part`, and the ones standing for data carry their identity too. `data-taskboard-id-key` prefixes the id with its type (`s:` / `n:`) because `1` and `"1"` are different cards and have to stay different keys, which a bare `String(id)` would merge; it is what `scrollToCard`, the keyboard and the drag hit test use to find an element back.

## Notable decisions

- **`collapsed` on a column seeds the state, it does not own it.** Only explicit user decisions are stored; the metadata is read as the default for a column nobody has touched. Re-reading it on every change would slam the columns shut under whoever had just opened them after a data refresh.
- **Read-only is not disabled.** Read-only means nothing can be CHANGED, which the runtime enforces on every write path; the cards stay content, so they keep their opacity, their text selection and any link inside them. Only `disabled` dims and inerts.
- **The two card shadows are separate rules, not composed variables.** Selected and focused say different things, so `selected`, `focused` and `selected.focused` are three explicit declarations — the styles package validator rejects declaring custom properties, and composing them would have needed one.
- **A column registers itself from its `column` INPUT, never from the resolved metadata.** Resolving falls back to the root's column list, which is the very thing the registration writes: reading it there makes the effect its own trigger and it spins forever. The registration also returns the same array when nothing changed, so the signal does not notify on a no-op.
- **A pointer drag still ends in a `click`.** Left alone that click changes the selection of the card just dropped, so a gesture that became a drag swallows exactly one following click.
- **Reordering is refused whole when it would displace a locked column.** `locked` promises a position, not merely that the column resists being picked up.
- **Renumbering touches only the affected cells.** Every untouched card keeps its own `order`, which is what a product persisting per-column ordering expects to see.
- **The board is not grouped until `swimlaneField` is set.** Passing `swimlanes` alone leaves the board ungrouped, because without the field it cannot tell which row a card belongs to.
- **The swimlane grid is authored by the application.** Angular content projection does not iterate, so the header row, the rows and the cells are written out with the public classes; the runtime keeps the grouping, the collapse state, the drop targets, the counts and the ARIA on its own wrappers.
- **A collapsed column keeps its body mounted and empty.** It is still the rectangle the pointer is tested against, and a column that vanished could not take focus.
- **The root carries a border.** It already had a `border-radius` with nothing to round, and every screenshot of the reference shows a bounded board; `taskboard.border.color` makes it a token, and the Scheduler in this repo is already put together the same way.
- **The add-card surface centres its content.** An add action at the foot of a list reads as a full-width target with its label in the middle, not a link hugging the left edge, so that is the default rather than something every consumer re-does.
- **Nothing measures geometry on the server.** Prerendering has an element tree and no layout engine, so a windowed column renders whole and starts windowing once there is something to measure.

## Not in it

`apps/docs/api-generator/typedoc.json` is not touched: it is generated by `build:apidoc`, and it is no longer tracked on `main`.

`apps/docs/public/demos.json` is not either: it is generated by `build:democode` and gitignored, so the per-example code blocks come from `pnpm build` through `build:docs:content`. Running `ng serve` against a stale file shows the examples with empty code blocks.

The repo's `lint` script does not currently match any files — `eslint.config.js` still carries the flat-config compat shape with `files: ['*.ts']`, which only matches root-level files, and the script's `--ignore-path` flag was removed in ESLint 9. Linting a file in this PR reports `File ignored because no matching configuration was supplied`. That is pre-existing and out of scope here; formatting is checked and clean.

## Docs

`/taskboard` with 32 runnable demos: the compound tree, data binding, custom identity fields, external mode, columns, column groups, swimlanes, card content, column content, UI parts, a product board header, selection, drag and drop, drop indicators, context menu, workflow rules, permissions, CRUD, undo and redo, keyboard, search and sort, custom predicates, virtual scroll, export and print, density and RTL, custom styles, projected content and contexts, events, data attributes, tree shaking, accessibility, performance.

The example data is the reference's own fixture verbatim — the same nine cards, descriptions and owners — so a reader comparing the two is comparing the component and not the copy. Search-and-sort, the predicates section and the board header all run OUTSIDE the board and hand it the result, because that is the boundary: the board has no opinion about what a card means, so it cannot know that critical outranks low, and a title with counters and actions is a product decision. The header derives its counters from `TASKBOARD_CONTEXT`, so they follow a drag with nothing telling them to.

## Tests

33 engine tests (the order the move rules fire in, the renumbering of source and target cells, cell-local range selection, external mode never writing, the collapse seed not overriding the user, a locked column refusing a whole reorder, CSV escaping, the snapshot round-trip and the virtual window) plus 25 component tests over the composed tree, covering the parts and their data attributes, definition-versus-projection fallback, the supplied visuals reading their contexts, collapse, selection, keyboard focus, the grouped container, the press-time selection guard, the preview clone, the loading surface and both print paths. The engine tests run without a DOM, on plain signals. Full suite green: 7336 passing.

## Summary by CodeRabbit

* **New Features**
  * Added the TaskBoard component for managing cards across columns, groups, and swimlanes.
  * Supports drag-and-drop, keyboard navigation, selection, collapse/expand, workflow validation, permissions, history, virtual scrolling, and custom card/column content.
  * Added card management, filtering, sorting, import/export, printing, state persistence, accessibility, and responsive behavior.
  * Added TaskBoard styling and design-token support for Aura, Lara, and Nora themes.

* **Documentation**
  * Added comprehensive TaskBoard documentation and interactive examples, accessible from a new documentation route.


nesgarbo:
Addressed all nine findings — every one was real. Pushed in `500de80`.

| Finding | What was wrong |
| --- | --- |
| Travelling cards not filtered for numeric ids | The drag snapshot stored the DOM's `data-task-id` (always a string) and compared it against the state's typed key: `s:7` vs `n:7`. The filter removed nothing, the dragged card kept counting towards the midpoint, and a drop below its original position resolved one slot too low. Reads the typed key the card already exposes. |
| Edge auto-scroll during a column reorder | The rAF guard tested the card gesture's flag, so the first frame returned and never rescheduled. |
| `addItem` ignoring `index` | Only the new card's `order` was written, so two cards held the same number and the stable sort put the newcomer second. The cells from the insertion point on now shift by one. |
| Repeated announcement silent | Alternating zero-width space, so the text differs without changing what is read out. |
| Read-only vs disabled cards | `inert()` covers disabled AND read-only, so every card on a read-only board was marked disabled: half opacity, no pointer, no text selection, no links. Read-only means nothing can be CHANGED, which the runtime already enforces. The board-level `pointer-events: none` went with it — it was taking the column scrolling too. |
| Date-only `dueDate` off by a day | Date-only strings are read as local days now, and the comparison is by calendar day, so a card due today is never overdue. |
| Custom marker not reactive | Measured after every render instead of inside a computed, so content appearing from an `@if` adds the class. |
| External-mode demo `order` collision | Renumbers the destination column, as the reorder handler already did. |
| `readonly` on a top-level variable | Moved into the class. |

Two more from the same round, found separately:

**Print lost the card spacing.** Flattening the column body to a block in the print sheet killed the flex gap the spacing comes from, so cards printed touching each other and the column edges. The body keeps its flex display now; only the padding, the surface and the scrolling go. The supplied parts also had no print treatment at all — the collapse chevron and count chip are hidden, the column title prints small, bold and black, and a tag keeps only its outline, because a tinted chip prints as a grey block with the label lost inside it.

**Text selection now has the real guard.** `user-select: none` covers the board's own surfaces, but a card projects arbitrary content and an application stylesheet can put `user-select: text` back on part of it. `selectstart` is cancelled while a press is tracked, which is the one hook CSS cannot override.

Comments, spec titles and fixtures are all in English now; some had been written in Spanish.

`pnpm test:unit` is green at 7336 specs (58 for TaskBoard), the package builds and `pnpm --filter docs build` are clean.

geromegrignon:

PR #1791, opened 2026-09-09
feat(texteditor): add text editor component

nesgarbo:
Adds a TextEditor component: `p-text-editor-root` owns the document, the commands, the state and the history, and every visible surface is a part the application fills with its own widgets. The editor ships no toolbar, no menu and no upload UI on purpose — each part renders an empty surface and hands its `commands`, `state` and request data to whatever is projected into it, through a `*...Def` slot template or through a context token. A part turns its feature on by being present in the tree: mount `p-text-editor-slash-menu` and `/` starts working, leave it out and it does not.

The implementation is original: the schema, the command set, the state derivation, the serializers, the sanitizer, the block-mode plumbing, the parts and the styles are written for this repo and follow the conventions the rest of optimus-ui already uses — signal state on a root, `*...Def` template definitions, `dt('component.*')` design tokens, `BaseEditableHolder` for the forms integration.

Deciding what an editor of this kind has to cover came from public material: PrimeUI Pro's published TextEditor documentation and demo pages, https://angular.primeuipro.dev/texteditor, which is where the API shape — the part list, the context tokens, the command set, the outputs — and the visual reference come from. No PrimeUI Pro source, package or bundle was consulted.

The document model is ProseMirror, the same engine the reference is built on, so `getView()`, `getState()`, `runCommand()` and `registerProseMirrorPlugin()` hand over the real thing rather than a lookalike. The twelve `prosemirror-*` packages are **optional peer dependencies**: an application that never mounts the editor does not install them, and nothing else in the library imports them.

Entry point is `@openng/optimus-ui/texteditor`, types live at `@openng/optimus-ui/types/texteditor`, tokens in the aura, lara and nora presets under `texteditor.*`.

## What is in it

**Parts** — 26 of them: root, toolbar, content, context toolbar and its overflow panel, block controls, block menu and submenu, slash menu, mention menu, the two upload overlays with their dropzone and progress halves, table controls, the column, row and cell menus with a submenu each, and the navigator with its trigger and menu. Every one carries `data-scope="texteditor"` and a `data-part`, which is the styling and testing contract; the renderless ones (navigator, table controls, uploads) carry none, because they have no DOM of their own.

**Two authoring styles** — 22 slot directives (`pTextEditorToolbarDef`, `pTextEditorBlockMenuDef`, …) hand over typed, already-unwrapped variables and are guarded by `ngTemplateContextGuard`, so `state.blod` fails the build. For content that needs its own class, project a component and inject one of the 12 tokens (`TEXT_EDITOR_CONTEXT`, `BLOCK_MENU_CONTEXT`, `MENTION_MENU_CONTEXT` — generic over the item type — and the rest), where the same fields arrive as signals.

**Two modes** — `classic` binds to an HTML string with a fixed toolbar above one content area; `block` binds to `string[]`, one entry per block, each draggable, each with a per-block options menu and a slash palette for inserting the next one. Markdown is a feature rather than a mode: `markdown` adds the input rules to either, live, without rebuilding the editor state or losing the undo history.

**Formatting** — bold, italic, underline, strikethrough, inline code, subscript, superscript and highlight, each with a command and a reactive state flag; headings 1–6 with `paragraph()` to convert back; bullet, ordered and check lists that convert between each other rather than nesting; blockquote and code block; alignment; text and highlight colour, font family and font size on one merged text style, so setting a size does not clear a colour.

**Content** — links that expand to the whole run around the caret when they are updated or removed, images from a URL or an upload, documents inserted as links, `@` mentions with a handler, a filter field and a template, and tables with column, row and cell menus, header toggling, merge and split, reordering, per-cell colour and alignment, and drag-to-resize columns.

**Uploads** — one queue per overlay: progress per file, cancellation through the `AbortSignal` the handler is given, client-side type/count/size validation that reports `reject` with a reason, `uploadError` when the transport rejects, and a single `complete` when the queue drains — which a cancelled queue does not fire. The overlay opens on `commands.uploadImages()` or on a drop, not by being mounted.

**Block mode** — a hover bar that follows the block under the pointer and stays reachable while the pointer crosses the gutter towards it, drag reordering with a drop indicator, a block menu with duplicate, copy, turn-into, colour and delete, and the slash palette. `Ctrl/Cmd + A` selects the block first and the document on the second press.

**Utilities** — the heading navigator (a minimap rail plus a popover outline that scrolls the document), undo and redo with availability flags, and a print command that renders the serialized content in a sandboxed frame without `allow-scripts`.

**Value and forms** — `[(value)]`, `[value]` or neither; `valueChangeDebounce` for large documents; `getHTML`, `getJSON`, `getBlocks`, `getText` and `getMarkdown`, all safe to call before mount. The root implements `ControlValueAccessor`, so `[formControl]` and `[(ngModel)]` work with no wrapper, `reset()` applies, `disable()` really blocks input, and setting `name` renders a hidden input so a plain `<form>` submit carries the content.

**Plugins** — `defineTextEditorPlugin(name, install, options?)`. The install function gets the selection, `replaceSelection`, the live view and state, `registerProseMirrorPlugin`, `runCommand` and an `onUnmounted` hook; the options argument contributes nodes and marks to the per-instance schema and ProseMirror plugins to the initial state. Commands land under `pluginCommands[name]`.

**Security** — everything entering the document is untrusted: parsed inside an inert `<template>`, filtered by the schema, and checked again on serialization. URL schemes are allowlisted (including obfuscated `java\tscript:` variants), `data:image/svg+xml` is refused while other inline images pass, an unsafe href drops the link and keeps the text, link targets are narrowed and `rel="noopener noreferrer"` is always emitted, and inline CSS is validated declaration by declaration.

**Accessibility** — the editing region is a real ARIA textbox that becomes a combobox with `aria-activedescendant` while a slash or mention menu is open and reverts when it closes; toolbars are `role="toolbar"` with pressed state on the toggles; menus share one popover primitive, so arrows, Enter and Escape behave the same everywhere and the active entry is marked rather than focused, because the caret has to stay in the document while the query is typed; upload progress is a live region; every animation is off under `prefers-reduced-motion`.

## Contract

The editor never resets itself: the ProseMirror instance is created once and survives its parents' change detection, so a re-render never costs the undo history or the caret. Writing back the value the editor just emitted is ignored, so the round trip does not loop, and a value set from outside is applied without adding a history entry — an external update cannot be undone into a state the application never had.

The client-side upload checks are a UX guard only. The browser reports the type and the size and both are spoofable, so the server has to check again; the docs say so where the inputs are documented.

## Notable decisions

- ProseMirror ships as optional peer dependencies rather than vendored or hidden behind a wrapper. A rich text editor has a floor, and pretending otherwise costs the plugin API its point.
- prosemirror-tables names its nodes in snake_case. They are renamed to camelCase at schema build, because the library finds them through `tableRole` rather than by name, and every command in this repo looks nodes up by a camelCase name.
- The checklist checkbox lives in a node view, not in `toDOM`. It is real, it is `contenteditable="false"` so Backspace cannot delete it, and it never reaches the serialized value: `getHTML()` emits `data-p-checked` and the checkbox is rebuilt from it.
- Placeholders are decorations. Text in the document would be selectable, copyable and serializable, which is three ways for a hint to end up in a database.
- ProseMirror dispatches outside Angular's event paths, so each transaction marks the editor for check. Without it, in a zoneless application, a selection-only change left every toolbar showing the previous selection's state.
- The style attribute is sanitized declaration by declaration. The browser normalizes a style attribute with a trailing `;`, and a check written for values rejected every colour and alignment the editor itself had just written.
- Block mode gives the content a gutter and the hover bar lives in it. With the bar outside the editor it was clipped, and hiding it when the pointer left the block made the handle vanish under the cursor on the way to it.
- Design tokens are `texteditor.*`, so the CSS variables are `--p-texteditor-*` rather than the reference's `--p-text-editor-*`. Component tokens in this repo are never hyphenated, and one component breaking that would be worse than the difference.
- `value` is a `model()`, so `valueChange` is the model's own output rather than a second one that would fire twice per edit.

## Not in it

- `apps/docs/api-generator/typedoc.json` is not touched: it is generated by `build:apidoc` and regenerating it buries the review.
- The reference's AI, translate and code-highlight plugin pages. They are the same plugin API pointed at a third-party service, and a demo of one is a demo of somebody's API key.
- Per-part secondary entry points (`.../parts/root`). Components in this repo have one entry point each, and the parts are standalone, so a template still references only what it imports.
- `onEditorKeyDown` / `onEditorMouseDown`, two internal DOM handlers the reference exposes as public methods. The equivalent behaviour lives in the plugins where ProseMirror already handles those events.

Two files outside the component are touched, both because this page is the first to hit them: `scripts/build-apidoc.mjs` crashed on a function type alias (a reflection with signatures and no children), and the docs section nav read `this.nav.nativeElement` from a timer that outlives the view.

## Docs

`/texteditor` with 49 sections and 43 runnable demos, following the reference's own page list: architecture, static and context toolbars, one section per inline format (bold, italic, underline, strikethrough, code, subscript, superscript), colours and fonts, one per structural command (heading, list, blockquote, code block, horizontal rule, text align), link, image, document, mention, table, block mode, block menu, slash commands, a complete block editor, markdown, navigator, undo/redo, print, events and output, controlled and uncontrolled, forms, readonly and disabled, the plugin API and seven example plugins (uppercase, typography, emoji, character count, focus, invisible characters and the schema-extending details block), slots and contexts, data attributes, security and accessibility.

The UI parts the demos use — the toolbar, the block and table menus, the slash palette, the mention list, the upload surfaces and the navigator — live in `apps/docs/components/texteditor` as ordinary application code, which is what they are.

## Tests

31 engine tests (URL and CSS sanitization including obfuscated schemes, HTML round-trips for every structure the toolbar produces, checklist state, markdown/text/blocks projections, the command set over a live view, format-state derivation, upload validation and formatting) and 13 component tests over the composed tree (mounting, the data attributes, the ARIA contract, the slot's unwrapped values, two-way binding, external writes, the serializers, disabled and readonly, the hidden input, block mode with reorder and drag-drop, and a form control driving and disabling the editor). 44 passing.

## Summary by CodeRabbit

* **New Features**
  * Added a rich text editor with formatting, lists, tables, uploads, mentions, slash commands, block editing, navigation, printing, plugins, Markdown support, and form integration.
  * Added customizable toolbars, menus, context toolbars, upload areas, and slot-based UI composition.
  * Added security sanitization, accessibility support, and light/dark theme styling.
* **Documentation**
  * Added comprehensive TextEditor documentation, examples, plugin demonstrations, and API guidance.
* **Bug Fixes**
  * Prevented navigation errors after documentation navigation is removed.
  * Improved API documentation generation for additional type shapes.


nesgarbo:
Pushed fixes for every review finding, plus a bug the review did not reach: **dragging a block did not land**.

## Block drag

The drop indicator is a decoration and decorations only redraw when the view updates, so `dragover` set the signal and nothing was ever painted — and the target came from the block under the pointer, which meant releasing over the lower half of the block above resolved to "leave it where it is". The indicator now refreshes on every `dragover` (`setProps({})`, as suggested on `block-mode.ts`), and the target is the nearest boundary measured against every block, so the gutter the hover bar lives in, the gaps between blocks and the space below the last one are all valid places to drop. The move is applied on `drop`, with `dragend` as the fallback for a drag that ends outside the content.

## Review findings

- `block-mode.ts` — the view is refreshed after the drag state changes, as above.
- `commands.ts` — `toggleList` and `toggleWrap` compose their lift and wrap into **one** transaction and route it through the provided `dispatch`, so probing no longer mutates and one undo no longer leaves a half-converted block. Covered by a test that counts dispatches.
- `keymap.ts` — `clearSelectedCells` maps both boundaries through the transaction.
- `keymap.ts` — the highlight shortcut goes through `setTextStyle`, so it merges instead of replacing the text style; the shortcut and the toolbar button now do the same thing.
- `tables.ts` — `clearContents` reads each cell from the transaction's own document and maps the positions.
- `texteditor-table.ts` / `texteditor-popover.ts` — the innermost open popover owns the keyboard: `entriesOf` no longer collects a submenu's entries as its parent's, and Escape closes the submenu without taking the parent with it.
- `texteditor.ts` — both pending timers are cleared in `onDestroy`.
- `input-rules.ts` — the italic rules keep the character before the delimiter and use `lastIndexOf`, so `t*t*` marks the right half.

Nitpicks: the upload class names are in `TextEditorClasses`, `aria-disabled` entries are skipped by popover navigation, and a plugin that throws during install no longer discards the cleanups the plugins before it registered.

## Other bugs found while testing this in the browser

- Anchored popovers stayed where they were opened while the page scrolled — the caret-anchored surfaces re-measure on scroll and resize, and the element-anchored ones re-read their anchor through a tick.
- The slash palette applied its command to the block under the pointer instead of the block the `/` was typed in.
- An uploaded document never appeared: the placeholder was removed and the link then inserted at a position where text is not allowed. Both overlays now turn the placeholder into the uploaded content in place.
- The floating toolbar could not be dismissed: a part that sets its own `display` beats the user agent's rule for `hidden`.
- Table headers carry `scope="col"`, and drag-to-resize is left out under RTL where the handle pulls from the wrong edge.

52 unit tests green, plus 40 end-to-end checks driven through the docs page (every format, the structural commands, markdown rules, tables and their menus, the block menu and drag, mentions, the slash palette, uploads with rejection, the navigator, forms, readonly and the seven example plugins).

nesgarbo:
Second round of review findings, all three looked at.

**`texteditor.ts` — cancel the block drag instead of applying the stale indicator.** Valid, and the reason `dragend` applied the move was that the gutter the hover bar lives in and the padding around the content are outside the content element, so the view's own `drop` never saw a release there. The component now listens for `dragover`/`drop` on its root as well, which makes those regions real drop targets, and `dragend` became a pure cancel: it clears the drag, clears the indicator, refreshes the decorations and emits `blockDragEnd` without moving anything. Escape mid-drag, or a release outside the window, leaves the document alone. A test covers the cancel; the existing test covers the drop.

**`commands.ts` — map the selection in the final coordinate space.** Valid. `transaction.selection` comes from the wrap transaction, so it is already in the merged document's coordinates, and `merged.mapping.invert()` was mapping it back toward the original. It re-resolves through an empty `Mapping` now.

**`package.json` — make the ProseMirror peers required.** Not applied; documented instead, which is the alternative the finding offers. `@openng/optimus-ui` is a single package covering every component, so a required peer would make twelve ProseMirror packages mandatory for an application that only uses, say, a button — the same reason `chart.js` is an optional peer here. The Import section of the TextEditor documentation now states that the packages are optional peers and gives the install command for the twelve of them, next to the existing note under Architecture.

53 unit tests green, library and docs build clean.

geromegrignon:

PR #1793, opened 2026-09-10
feat(charts): add charts component

nesgarbo:
Adds a Charts component: `p-chart-svg` and `p-chart-canvas` own the data, the scales, the layout and the hover state, and every child part reads what it needs from that shared state instead of receiving a dozen inputs from a parent that would have to forward them. A series exists because it is declared inside the root — `<p-chart-bar [data]="…">` — a feature exists because its element is in the tree, and any repeated surface can be swapped for a `*pChart…Def` template. A part *is* its own configuration: there is no options object anywhere.

The implementation is original: the scale and tick engine, the arc and path geometry, the stacking, the decimation, the treemap tiling, the label layout, the hit testing, the export writers, the parts and the styles are written for this repo, and they follow the conventions the rest of optimus-ui already uses — signal state on a root, `*…Def` template definitions, `dt('component.*')` design tokens.

Deciding what a chart library of this kind has to cover came from public material: PrimeUI Pro's published charts documentation and demo pages, https://angular.primeuipro.dev/charts, which is where the feature list — the chart families, the part set, the feature surface — and the visual reference come from. No PrimeUI Pro source, package or bundle was consulted.

Entry point is `@openng/optimus-ui/charts`, styles in `@openng/optimus-ui-styles/charts`; tokens live in the aura, lara and nora presets under `charts.*`.

**This is an addition, not a replacement.** The existing chart.js-backed `p-chart` (`ChartModule`, `ChartStyle`) is untouched and stays: it is a thin wrapper over a library an application may already be standardised on. `ChartsModule` is the alternative for the cases that wrapper cannot serve — declarative composition, a shared scene across two renderers, and a chart that can be styled from CSS.

## What is in it

**Two renderers over one scene.** Both roots build the same list of `SvgNode` descriptors; the SVG root materializes them as elements and the Canvas root replays them with a painter. Geometry is therefore identical by construction rather than by careful duplication, and switching renderer is a tag change. SVG is the default because every mark is a real element — styleable from CSS, readable by a screen reader, findable by a test, printable at any resolution. Canvas is the trade you make when a hundred thousand marks would mean a hundred thousand nodes.

**Ten chart families** — column and bar, line and area, pie and donut, scatter and bubble, radar, polar, heatmap, treemap, candlestick and OHLC, and combos of any of them in one root. Nine series parts (`p-chart-bar|line|pie|scatter|radar|polar|heatmap|treemap|candlestick`) plus five grouping parts (`p-chart-stacked|overlap|waterfall|range|treemap-group`) that change how their children resolve rather than drawing anything themselves.

**Scales** — band, linear, logarithmic and time, with nice ticks chosen from the axis length rather than a fixed count, `startFromZero`/`softMin`/`softMax`, several axes per chart through `xAxisId`/`yAxisId`, minor grids, alternating bands, auto-rotation and auto-skip when labels would collide, and a second row of labels from `p-chart-axis-group` when twelve months are really four quarters. Membership in a group can be an array or nested `p-chart-axis-category` children, because both read well in different situations.

**Arc geometry is not angular padding.** A slice gap set in angle is wide at the rim and closed at the apex, which is exactly the nightingale defect the reference does not have: `paddedArcPath` offsets each edge perpendicular by a constant width and truncates the apex at `(w/2)/sin(sweep/2)`, so the gap measures the same everywhere and the centre keeps a hole. `borderRadius` on a slice or a ring is real tangent corner arcs, capped by half the ring thickness and half the arc length at each edge.

**Hover resolves by containment, not by x distance.** Each family answers "what is under the pointer" the way its own marks are shaped — rect containment for heatmap cells, treemap cells and bars, the x band for a candle, euclidean distance for a scatter point, x then y as a tiebreak for a line. A candidate inside a mark always beats a nearer candidate outside one, which is what makes a combo highlight the series you are actually over. `p-chart-hover` then decides what highlighting looks like: brightness, dimming of the rest, offset, scale, marker radius, an explicit colour, or a border.

**Tooltip** — `mode` (item, index, dataset), snapping, a crosshair, follow-cursor or anchored placement, show and hide delays, a value formatter, and `*pChartTooltipDef` for the whole surface. It is shared across a `p-chart-group`, so a dashboard reads one instant across four charts.

**Legend and colour legend** — position, align, layout, interactive toggling, icon shape and size, `*pChartLegendItemDef`; and for the continuous scales a `p-chart-color-legend` with its own ticks, steps, indicator and label formatter. Palette slots are shared with the series painters, so a legend swatch and its marks cannot disagree.

**Data labels** are one collision-resolving layout pass over every label the chart wants, not a per-series decision made in isolation. Placement is per family; pie and donut leader lines elbow on one ring at `radius + distance + textGap`, so a dozen labels line up instead of each finding its own elbow. Label colour is picked by measured WCAG contrast ratio against the mark it sits on, not by a luminance threshold — a mid-tone fill is where a threshold puts white text on a light background.

**References and annotations** — `p-chart-reference-line` and `p-chart-reference-band` in data coordinates on either axis, with label placement, background and styling; `p-chart-annotation` with `*pChartAnnotationDef` for anything a page wants to draw in the plot, positioned at data values and following zoom.

**Zoom, pan and navigator.** Zoom writes a window to the chart state and stops there: the scales are rebuilt from the narrower domain and the same painters draw it, so text stays crisp and the ticks re-nice themselves — a transform could do neither. Wheel, drag, pan, pinch, modifier keys, limits, a reset button and stepper buttons. `p-chart-navigator` draws its own mini chart of the *full* domain under the plot with a draggable window, because once you have zoomed in, zoom alone cannot tell you where you are.

**Decimation** — LTTB, min-max and k-means, with a sample count or a threshold. The rows are untouched and each kept point carries its original index, so a tooltip still finds the row it came from; removing the element gives back every point without touching the data source.

**Export** — PNG, JPEG, SVG and CSV, plus transparent PNG and SVG, and PDF through a hand-written PDF-1.4 writer rather than a dependency. `p-chart-export-menu` with `*pChartExportMenuItemDef`/`*pChartExportMenuIconDef`, or `exportChart()` on the root imperatively. The CSV and the accessible data table are generated from the same derivation, so a spreadsheet and a screen reader read the same chart.

**Accessibility is on by default, not opt-in.** The root always publishes a description and a data table, visually hidden with `clip-path: inset(50%)` so they stay in the accessibility tree without taking space. `p-chart-accessibility` does not switch that on — it registers the options that shape it: prose, type description, heading level, row cap, cell and series formatters, landmark verbosity, point-description threshold, keyboard navigation, and colour-vision patterns for the cases colour alone carries the meaning.

**Responsive** — rules evaluated against the chart's *own measured box*, not the viewport, because two charts side by side on a wide screen are both narrow and a media query cannot tell. Auto-adaptive scaling (title, caption, axis labels following a size tier) is on without the element; the element adds the rules, breakpoint overrides, and the ability to switch adaptation off.

**Multi-chart sync** — `p-chart-group` links crosshair position, zoom range and series visibility. It takes no inputs of its own: `sync` on each root decides what participates, because one chart in a dashboard may follow the shared zoom while another shows a fixed reference range. A legend placed in the group but outside the roots controls all of them.

**Treemap drilldown** — squarify tiling with hierarchy from `parentField`/`nodeId`, group padding and labels, per-level styling, `interactByLeaf`, and `p-chart-breadcrumb` as the trail back up. The breadcrumb is real DOM because every level in it is a link, and it lives beside the legend and the title rather than inside the treemap, which draws cells.

**Plugins** — `defineChartPlugin` gives an extension seam: a plugin reads chart state, subscribes to frames and hover, paints an overlay above the chart and can publish its own API. What it cannot do is change how the chart draws itself, which is the point — no future version has to preserve internals for it.

**Internationalization** — `locale`, `numberFormat` and per-axis date formats all go through `Intl`; `dir` on the root plus logical properties flip the plot, the axes, the legend and the chrome together. Fourteen language catalogues ship from `@openng/optimus-ui/charts/locales` as separate exports, so an application that needs German does not carry the other thirteen; English is built in, because a chart with no catalogue registered still has to be able to describe itself. Every catalogue except English is machine-drafted and unreviewed, and says so.

**Chrome and text** — `p-chart-title` and `p-chart-caption` with position, alignment, floating and full type control; `p-chart-item` for declaring points inline instead of binding an array; `animation`/`animations`/`transitions` per property with custom easing; `theme`, `fontFamily`, `fontSize`, `aspectRatio`, `scaleWithRootFont`, `debounceDelay`; `redraw()`, `updateData()`, `toDataURL()` and `exportChart()` imperatively, and `(pointHover)` on the root.

## Contract

The component never mutates the bound `data`. Decimation, stacking, sorting and drilldown all derive new arrays; the rows an application passed in come back untouched, and a hover payload or a CSV row carries the original index so the application can find its own record.

A part *is* its own configuration. Parts register on mount and deregister on destroy, and the root paints from that registry — children draw nothing themselves. Removing `<p-chart-decimation>` gives back every point; removing `<p-chart-hover>` gives back the default highlight. There is no options object to keep in sync with the tree.

## Notable decisions

- One shared scene, painted twice. The alternative — an SVG painter and a Canvas painter written against the same layout — drifts, and it drifts in the direction of whichever one is looked at more.
- Slice gaps are a width, not an angle. Angular padding is the reason nightingale wedges converge at the centre and fan open at the rim.
- Hit testing is per family, with `inside` beating `nearer`. An x-distance-only test is why a combo highlighted the wrong series and a heatmap highlighted no cell at all.
- Label colour comes from a measured contrast ratio. A luminance threshold picks white text on a mid-tone fill, which is legible in neither direction.
- The accessible description and the data table live in one component, because they come from one derivation: the table is the numbers and the description is a summary of the same numbers. Deriving them apart would eventually let the sentence disagree with the table under it.
- Font sizes scale as `round(base × rootSize / 16)`. Measured against the reference across five root sizes, not assumed.
- The engine (scales, ticks, stacking, curves, decimation, treemap tiling, geometry, formatting) is plain functions over plain data, tested without a DOM.
- Zoom changes the domain, never a transform. A scaled copy blurs text and keeps the old tick values.
- The navigator draws its own SVG rather than going through the scene: it has its own vertical scale and always the full domain, and it must keep showing the whole series while the chart's scales show a slice.
- Nothing measures geometry on the server. Prerendering has an element tree but no layout engine.

## Not in it

`apps/docs/api-generator/typedoc.json` is not touched here: it is generated by `build:apidoc` and regenerating it produces a ~860k-line diff that would bury the review. #1782 takes it out of git for good.

Treemap tiling divides differently from the reference in some hierarchies. Both are valid squarify and the areas are correct; the split order is not pinned.

## Docs

`/charts` — one page, in this repo's shape, 415 sections across 15 groups: getting started, column and bar, line and area, pie and donut, scatter and bubble, radar, polar, heatmap, treemap, candlestick and OHLC, combo, synced charts, configuration, internationalization, reference. 318 of them are runnable demos, deferred `on viewport` — without that, a page that mounts several hundred charts at once, some with a hundred thousand points and live timers, never finishes loading.

Configuration is the large group (115 sections): animation, transitions, annotations, axes, grids and ticks, hover, tooltip, legend, colour legend, data labels, reference lines and bands, zoom, navigator, export, responsive, decimation, plugins, passthrough and data attributes. Reference covers accessibility, performance with runnable large-data evidence, SVG versus Canvas, and the API tables.

Demo prose and demo data are the reference's own, not paraphrases. Demo source is served per component instead of one catalogue: charts alone is 624 kB of demo code, and every other page was downloading it.

39 of the 40 part selectors appear in at least one demo — the internal `p-chart-a11y-view` is the exception, since the root mounts it. Nine of the fifteen template definitions are demonstrated; `pChartAxisGroupDef`, `pChartCenterContentDef`, `pChartColorLegendDef`, `pChartExportMenuIconDef`, `pChartExportMenuItemDef` and `pChartReferenceLineDef` are documented in the API tables but have no demo — the reference demonstrates none of those six either. Four root inputs (`aspectRatio`, `debounceDelay`, `scaleWithRootFont`, `transitions`) and the `(pointHover)` output are likewise documented but not demonstrated. Checked mechanically against the component's own API rather than by eye.

## Tests

72 engine tests (band, linear, log and time scales, nice tick selection, monotone-cubic curves, stacking and waterfall running totals, LTTB/min-max/k-means decimation, squarify tiling, arc geometry including the constant-width gap and the corner-radius caps, contrast ratio, number and date formatting, CSV) plus 38 scene tests over the composed tree, including that the SVG and Canvas roots agree on the scene, that hit testing lands inside the right mark per family, and that the accessible table matches the exported CSV. Full suite green: 7388 passing, 90 skipped.

Appearance was verified by measurement rather than by eye: per-class computed-style diffs against the reference, scripted hover probes at known fractions, and rAF counting to confirm an idle chart is idle. That is what turned up the tick label colour, the axis title weight and size, the annotation colour, the `p-chart-tick` class name, the radar spoke colour, the radial rim divisions, and the heatmap row order.


nesgarbo:
@coderabbitai review

devfake:
WHO tf is gonna review your massive pull requests?

geromegrignon:

PR #1799, opened 2026-09-11
feat(docs): rebuild the documentation site on Analog

nesgarbo:
The documentation site, rebuilt on Analog and prerendered to 121 static pages, with the look of the 2026 design and the measurements to go with it.

Everything here is `apps/docs`, plus the linter setup and a dependency bump. The library is untouched.

## Working in it

Adding or changing a page is the same work it was, minus the plumbing. A component page is still its demo in `src/demos/<name>/index.ts`, its sections in `src/doc/<name>/*-doc.ts`, and its entry in `menu.json`; `<app-code>` still takes no inputs and reads the snippet extracted from the demo's own source.

What is gone is the routing. There is no `pages/<name>/routes.ts` to write, no entry to add to `router/app.routes.ts`, and no `build:routes` to remember: the route is the one-line page file that re-exports the demo,

```ts
// src/app/pages/(docs)/badge.page.ts
export { BadgeDemo as default } from '@/demos/badge';
```

and the prerender list is derived from the same menu the navigation reads, so a page cannot be prerendered and missing from the menu, or the other way round. The generators keep their names and their behaviour (`build:apidoc`, `build:democode`, `build:llm-docs`, `build:sitemap`), with three added for the changelog, the component index and the open-issue counts.

## What it looks like

- **Topbar** — one 3.5rem row, the mark at the start and the controls at the end, inside a frame capped per breakpoint so a 1920px window keeps its gutters instead of stretching to the edges. The footer and the reading column are capped from the same numbers.
- **Navigation** — three sections behind a switcher with only the active tree rendered, the deepest matching item selected, and a discreet version selector at the foot of the rail.
- **Component pages** — the tabs above the page are gone. One column of prose and examples, a breadcrumb over the title, and a right-hand rail listing what the page has: Features, API, Theming, Passthrough, Known Issues, Copy Markdown, then an index of the page's own sections. The Features count is built from the sections that exist rather than kept by hand.
- **Landing** — rewritten as a sequence of sections that each say one thing: a hero with six application samples, what the library composes into, design tokens edited in place, accessibility, developer experience, the stack the site itself is built on, and the community numbers read live from npm. Nothing on it claims a customer we do not have: where the reference shows borrowed enterprise logos, this shows what the project is — MIT, community maintained, and the downloads it really gets.
- **Two pages that were missing** — an index of every component with a live preview of each, and a changelog of every release.
- **Search of our own** — results grouped by page, the typed text marked in each hit, arrow keys, recent searches, and an index built from the menu and the component list so it answers even while the Algolia application is blocked.
- **Theme designer** in a drawer from the topbar, an **RTL switch** in the configurator, and the **assistant** as a launcher in the corner, styled to belong to the site in both themes.
- **Syntax-highlighted code** in both themes from one pass, where the previous site served it as plain text.
- **On a phone** — the samples reflow instead of panning, the navigation opens from a button that nothing covers, and the section index becomes an island that animates out to the full width when opened.

## What it measures

Same machine, same network, both served with gzip; "before" is production as it stands today.

| | Before | After |
|---|---|---|
| FCP / LCP `/button` | 432 / 452 ms | **48 / 48 ms** |
| Page weight `/button` | 3561 KB | **1273 KB** |
| Third-party origins | 7 on every page | **0** (one on the home page, for the npm count) |
| CLS `/button` | 0.003 | **0** |
| DOM elements `/table` | 2592 | 2551 |
| `og:title` / `og:description` | the site's own on every page | **121/121** correct |
| JSON-LD | none | **121/121** |
| Code in the prerendered HTML | 2 blocks on `/button` | **19** |
| Pinch zoom on a phone | disabled | allowed |
| `/llms.txt` | 200 with the not-found page | the file |
| Images without alt text | 5 on the home page, 2 per component page | **0** in our own markup |
| Meta descriptions | 7 too short, 2 too long | all **72–165** characters |
| `pnpm run lint` | did not run at all | **0 errors** |

Titles, descriptions and canonicals are unique and self-referencing on all 121 pages; one `h1` on 120 of them, the exception being the not-found page, which is `noindex` and no longer listed in the sitemap. Of the 118 URLs in the previous sitemap, **none is lost**.

## The interesting fixes

- **Hydration was never enabled**, so the browser threw the prerendered HTML away and rebuilt the page — 0.108 of layout shift on a component page and the markup replaced before anyone could interact with it. Turning it on also needed the server to say it is the server: Angular reads a global the CLI writes into its own server bundle, and with nothing writing it here the framework took the prerender for a browser and reached for `window`.
- **The examples' code was fetched in the browser**, 168KB a visit, and arrived a few hundred milliseconds late — doubling the page height under the reader and leaving the prerendered HTML, which is what a crawler and every language model read, with no code in it.
- **The assistant loaded on every page**: two megabytes of widget plus a megabyte of reCAPTCHA, for a panel most readers never open. It loads on the first click now.
- **Highlighting put a span around every token.** They are compacted — a token in the body colour needs no element, two neighbours painted alike need one — and a block is coloured when the reader comes within a screen of it. Colour changes nothing about layout, so nothing moves: `/table` loads 2551 elements where it loaded 15,129.
- **Fonts came from three third-party origins**, one of them a personal site, each blocking the first paint. They are ours now.
- `eslint.config.js` was still in the `.eslintrc` shape, which eslint 9 does not read: the lint script failed before linting a file. It is a flat config; the library's older idioms report as warnings, the documentation passes the full set. It surfaced real bugs on the way — two `typeof window !== undefined` comparisons that were never true, an icon list sorted by a doubled comparison, and a reduce that mutated its accumulator.

## Checks

`pnpm run lint` 0 errors · `format:check` clean · `build:docs` 802 demos with 0 errors, 5/5 extraction suites, 801 StackBlitz projects · `test:styled` 25/25 · `test:unit` 7278 passed · production build: 121 pages, critical CSS inlined in all of them.

Exercised in a browser: ⌘K search with keyboard navigation, theme toggle and its persistence, the configurator's primary and RTL, the designer drawer, all five rail panels, code copy and StackBlitz, the version selector, menu navigation, anchors, the components index and its filter, the changelog, the assistant, and on a phone the navigation drawer and the section island. No page scrolls sideways at 390, 768 or 1280.

## Known, and left alone

- `p-avatar` renders its image with no `alt` (11 on `/avatar`, 3 on `/accordion`). That is the library, which this branch does not touch.
- Ninety-nine requests on a component page, against forty-eight before: one chunk per lazy component, already merged below a 30KB floor.
- ESLint 10, vitest 5 and pnpm 12 are held back deliberately; the reasons are in the dependency commit.


geromegrignon:
Closed due to:

- no design discussion in an issue
- making off-topic changes

PR #1800, opened 2026-09-11
feat(aiassistview): add the AIAssistView component

nesgarbo:
Adds an AIAssistView component: `p-aiassistview` owns the transcript, the composer, the streaming lifecycle, the reasoning and tool rendering, attachments, speech, the keyboard and the accessibility contract, and every supplied part reads that shared state and calls the same commands the root does instead of receiving inputs from a parent that would have to forward them. That is what keeps a replacement part honest: it can only do what the root already does.

What the component does **not** own is the model. Every answer arrives through `promptRequest` and goes back through the handle, because there is no shape of AI back end this component could sensibly assume.

The implementation is original: the state, the block grammar, the markdown renderer, the stream normaliser, the parts and the styles are written for this repo and follow the conventions optimus-ui already uses — signal state on a root, `*...Def` template definitions, `dt('component.*')` design tokens.

Deciding what an assistant of this kind has to cover came from published material: Syncfusion's AI AssistView documentation and demo pages, https://ej2.syncfusion.com/angular/demos/#/bootstrap5/ai-assistview/, and the `.d.ts` type declarations published with `@syncfusion/ej2-interactive-chat` on npm, which is where the feature list comes from. Those declarations are signatures and doc comments; no Syncfusion implementation, bundle or stylesheet was used, and none of their API names, CSS classes or tokens appear here.

Entry point is `@openng/optimus-ui/aiassistview`; tokens live in the aura, lara and nora presets under `aiassistview.*`.

## Composed out of the library, not restated

The header is a `p-toolbar`, the view switcher `p-tabs`, the boxed blocks `p-panel` (the reasoning one `[toggleable]`), the composer a `p-inputgroup` around a `pTextarea`, the errors `p-message`, and the buttons, chips, avatars and progress bars are the same ones every other page uses. Icons come from the OpenNG set through `OpenngIcons` — **no icon was added to the icon package**.

The practical effect is a small stylesheet: 72 design tokens, almost all resolving to semantic ones (`{content.*}`, `{text.*}`, `{primary.*}`, `{form.field.*}`), so the assistant follows a re-themed preset without a second pass. What is left in the component's own CSS is the layout, the transcript typography and the three surfaces it genuinely invents.

## What is in it

**Streaming through one call.** `streamResponse(source, options)` takes a `fetch` `Response`, a `ReadableStream`, an SDK's async iterable or an `Observable`, and each delta is appended as it lands. `sse: true` re-frames the source as Server-Sent Events and `transform` picks the delta out — a JSON string from a hosted model, or the object an SDK already parsed. The stop button cancels the read and releases the connection, a stream that throws marks the turn failed with a retryable error block instead of leaving it open, and bytes are decoded with `stream: true` so a character split across two chunks survives.

**SSE frames are reassembled, not assumed.** A frame ends at a blank line and a chunk boundary lands mid-frame constantly; treating each chunk as a message loses half the answer. The carry-over buffer is what makes it correct, and the `[DONE]` sentinel is swallowed rather than delivered.

**Structured answers.** A response is a list of blocks — `text`, `code`, `thinking`, `tool`, `error` — because a modern assistant interleaves prose, reasoning, tool calls and code, and each needs affordances a single rendered string cannot offer. A plain-string answer is split into the same shape, so a back end that returns markdown gets copy buttons and language labels for free.

**Reasoning timelines.** A `thinking` block is stages with a status, a line of text and context chips — files read, searches run, tools called — each with its own outcome badge. `updateBlock` patches it in place by id while the model is still reasoning, and the panel counts elapsed time while `isActive`. A clickable chip reports through `contextClick`, which is how an editor integration opens the file the model looked at.

**Templates.** A `*Def` directive behind every surface — prompt, response, block, tool, stage, suggestion, view, toolbar item, attachment, banner, footer, header — each typed through `ngTemplateContextGuard`, so a `let-` binding is a real type and a misspelled field fails the build. Syncfusion's equivalents are `string | Function` inputs, which give up type checking, change detection and content projection at once.

**Attachments** by picker, drop anywhere on the surface, or paste. Accept list, size and count rules, with a rejection announced through the assertive live region rather than failing silently. Left without a `saveUrl` the files stay local with their `File` intact, which is what most assistants want; set one and the component uploads and reports progress per chip.

**Speech** over the browser's own engines. Both APIs are optional in the platform and absent on the server, so the microphone is drawn only where it would work and read-aloud joins the response toolbar only when a voice exists. A silence ends a dictation, because continuous recognition never ends itself.

**Regeneration keeps the previous answer.** A rerun pushes what was on screen into `regeneratedResponses` and the reader steps between them. A regeneration that throws the old answer away is the one thing every reader complains about.

**Store-owned mode.** `managed` off and the component stops writing entirely: it renders `prompts` and emits its outputs as requests. Every cancellable payload still works, so a store can refuse a send, an edit or a clear before the surface has acted on it.

## Contract

The component never mutates `prompts` behind the application's back: every write replaces the array, so `OnPush` sees the change and a consumer holding the same reference is not surprised. `promptRequest`, `promptEdit`, `suggestionClick` and `clearRequest` are cancellable, and a cancelled request leaves the transcript and the composer exactly as they were.

`toolbarItemClick` fires **before** the built-in behaviour of a known entry, so setting `cancel` puts your own behaviour in its place — and it fires for ids the component does not know, which is how a product's own entries are delivered.

## Notable decisions

- The markdown subset is written here rather than pulled in. What a chat answer uses is small and closed, and a general converter brings a parser, an HTML passthrough mode and a sanitiser question with it. Every input is HTML-escaped before a single rule runs, `javascript:` hrefs are refused before they can become a valid anchor, and the result still goes through Angular's sanitiser on `[innerHTML]`.
- A turn, not a message, is the unit of the transcript. Prompt and answer are edited, regenerated and navigated together.
- The reader's fold of a reasoning panel is kept in the shared state, keyed by block id — otherwise the panel springs back open on the next streamed patch.
- The transcript follows new content only while the reader is already at the end. A stream that drags them away from something they scrolled back to read is worse than one that does not follow at all.
- `flat` is the default layout, not `bubble`. A tinted lozenge fights every panel, table and code sample the answer itself contains, and there is no bubble anywhere else in the library to match. `bubble` is there for products that want the chat shape.
- The composer's autosize sets an explicit height **only** once the content needs more than one row. An explicit height opts the field out of the input group's stretch and leaves it a pixel short of the addons beside it.
- Ids are a prefixed counter, not `crypto.randomUUID`. They are generated during rendering and a random id differs between the server render and the client one, which throws the hydration away.
- Reasoning is left out of copy, export and read-aloud. It is scaffolding, and pasting it into a document is never what was asked for.

## Docs

`/aiassistview` with 17 runnable demos: import, basic, streaming, response blocks, reasoning, tool calls, suggestions, toolbars, attachments, speech, views, templates, composition, store-owned mode, layout and RTL, events, accessibility.

The fixture answers from a canned set rather than a key, and streams through the same async generators a provider's SDK would hand over — including a source framed as SSE with frames cut mid-chunk, so the demo exercises the reassembly rather than a happy path.

## Tests

29 tests: 15 over the composed tree (a prompt appending a turn and clearing the composer, a cancelled request changing nothing, a streamed answer accumulating and closing, blocks patching by id, regeneration keeping the earlier answer and the navigator counting it, feedback toggling off, export) and 14 over the markdown and block splitting, where a mistake is a security bug rather than a layout bug — escaping, `javascript:` links, code spans, tables, and the half-written fence a stream spends most of its time in.

Five of the component tests cover the stream path specifically: plain deltas, SSE frames cut mid-frame, an SDK yielding objects, a stream that throws, and stop pressed mid-read. Full suite green: 97 files passing.

## Summary by CodeRabbit

* **New Features**
  * Added the AIAssistView conversational assistant component with prompts, streaming responses, markdown, reasoning, tools, suggestions, attachments, speech, custom views, toolbars, and transcript export.
  * Added configurable accessibility, keyboard navigation, responsive layouts, compact mode, and light/dark theme support.
  * Added Aura, Lara, and Nora styling presets.
* **Documentation**
  * Added a dedicated AIAssistView documentation section with interactive examples covering core features, customization, events, templates, streaming, attachments, speech, and accessibility.
* **Tests**
  * Added coverage for rendering, streaming, markdown safety, transcript actions, and response handling.


geromegrignon:

nesgarbo:
Pushed three commits: the four API gaps against the reference component, the review findings, and the docs.

**A bug the demo caught.** `promptRequest` went out *before* the turn was opened, so a handler that answered synchronously — `streamResponse(...)` inside the handler, which is how the streaming demo is written — found no open turn and did nothing. The demos that answered inside a `setTimeout` hid it. The turn is opened first now, and rolled back in full when a handler cancels.

**Gaps closed** (checked against the published type declarations, not from memory): `showClearButton`, `pAssistTurnDef` (the whole-turn template), `registerToolUI`/`unregisterToolUI`, and `promptToolbarSettings`/`responseToolbarSettings` taking the same options object the header and footer toolbars already took.

Writing the tests for those turned up two more defects: content queries were not reaching into embedded views, so a definition inside an `@if` was never found; and the projection point only existed inside the custom-view branch, so projected content that is never projected was never instantiated and such a definition did not exist to be found either.

**Review findings** — all addressed. Two are worth calling out:

- The generated `href` was exposed to the emphasis rules running after it, so `https://host/my_file_latest` arrived with an `<em>` inside the URL. Everything the renderer emits is now lifted out of the text and put back at the end, while a link's visible label stays exposed so `[**bold**](url)` still works.
- Remote images are no longer fetched by default. Model output can be talked into ``, which leaks the reader's address as soon as the answer renders. `allowRemoteImages` turns them back on.

**One finding I did not fix here**, because it is not this component's: `p-panel` puts `aria-expanded` and `aria-controls` on the `<p-button>` host rather than on the `<button>` inside it, so a screen reader reads the toggle without its state. It affects every toggleable panel, not just the reasoning block, and belongs in its own change. The accessibility page no longer promises behaviour that depends on it. Happy to open that separately.

Suite green: 97 files, 24 tests over this component.

